PR #5, opened 2016-09-03
Navio: Add support for SPI-connected GPS

HidenoriKobayashi:
This adds support for a U-blox GPS device on SPI bus. This needs another PR for the Firmware.

### DIFF
--- a/src/ubx.cpp
+++ b/src/ubx.cpp
@@ -77,7 +77,7 @@
 #define UBX_WARN(...)		{GPS_WARN(__VA_ARGS__);}
 #define UBX_DEBUG(...)		{/*GPS_WARN(__VA_ARGS__);*/}
 
-GPSDriverUBX::GPSDriverUBX(Interface interface, GPSCallbackPtr callback, void *callback_user,
+GPSDriverUBX::GPSDriverUBX(gps_interface interface, GPSCallbackPtr callback, void *callback_user,
 			   struct vehicle_gps_position_s *gps_position,
 			   struct satellite_info_s *satellite_info) :
 	GPSHelper(callback, callback_user),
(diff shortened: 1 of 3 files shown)


src/ubx.cpp
```diff
@@ -108,10 +108,12 @@ GPSDriverUBX::configure(unsigned &baudrate, OutputMode output_mode)
 {
 	_configured = false;
 	_output_mode = output_mode;
+#if !defined(__PX4_POSIX_RPI)
```
LorenzMeier: It would be good if you could change this to an interface flag instead ```configure(enum interface, unsigned baud, output mode)```. Then the interface setting controls wether or not the baud rate is applied

HidenoriKobayashi: I think it would be better if we add the interface information to the GPSDriverUBX constructor since the interface information is static?

LorenzMeier: Makes sense, yes. That's better than my original proposal.

src/ubx.cpp
```diff
@@ -413,6 +436,18 @@ GPSDriverUBX::receive(unsigned timeout)
 				handled |= parseChar(buf[i]);
 				//UBX_DEBUG("parsed %d: 0x%x", i, buf[i]);
 			}
+
+#if defined(__PX4_POSIX_RPI)
```
LorenzMeier: Why is this RPI specific? Isn't this rather SPI interface specific?

LorenzMeier: So should use ```if (_interface == GPSHelper::Interface::SPI)``` instead.

LorenzMeier: @mayanez Have you seen this bit of feedback?

bkueng:
cleaned up & applied

PR #12, opened 2016-12-26
Untested: UBX driver: Reset module as part of config

LorenzMeier:
@bkueng @sirPerna This should fix the issues with pre-configured GPS modules which do not accept serial config commands. I don't have one of these modules so I need to test this first. Matching Firmware PR:

https://github.com/PX4/Firmware/pull/6171


AdamSiwy:
@LorenzMeier  I would really test the UBX Reset on the M8N GPS module. could you give me a nudge how to implement the submodule in PX4 Firmware & QGC. Greetings

LorenzMeier:
@AdamSiwy Sorry, I never saw this comment. I'm now coming back to it.

PR #18, opened 2017-02-05
c++11 initialization cleanup

dagar:


src/ashtech.h
```diff
@@ -72,16 +72,16 @@ class GPSDriverAshtech : public GPSHelper
 		NME_DECODE_GOT_FIRST_CS_BYTE
 	};
 
-	struct satellite_info_s *_satellite_info;
-	struct vehicle_gps_position_s *_gps_position;
-	uint64_t _last_timestamp_time;
-	int _ashtechlog_fd;
+	struct satellite_info_s *_satellite_info {nullptr};
```
bkueng: @dagar I prefer the  `= nullptr` intialization, because it is more consistent with what we already have & are doing. What's your take? Is this actually the prefered way of doing it in C++11?

dagar: Yes, that would have initially been my choice as well. 
It seems to be the preference in c++ core guidelines. It's definitely what the tools want to do by default. Let me see if it actually says anything.

dagar: It sounds like uniform initialization is the way forward. http://www.stroustrup.com/C++11FAQ.html#uniform-init
What do you think?

Perhaps we should establish some basic coding standards to gradually bring more consistency?

bkueng: I'm all for brace-initializer then. Looks like it's better, also in terms of type safety.

Yes some basic standards would help, I'd just avoid having too many rules that make it hard to bring in any changes at all.

dagar:
@bkueng good to go?

PR #25, opened 2017-06-25
ublox allow setting dynModel

dagar:
https://github.com/PX4/Firmware/pull/7470

### DIFF
--- a/src/ubx.h
+++ b/src/ubx.h
@@ -160,8 +160,6 @@
 
 /* TX CFG-NAV5 message contents */
 #define UBX_TX_CFG_NAV5_MASK		0x0005		/**< Only update dynamic model and fix mode */
-#define UBX_TX_CFG_NAV5_DYNMODEL	7		/**< 0 Portable, 2 Stationary, 3 Pedestrian, 4 Automotive, 5 Sea, 6 Airborne <1g, 7 Airborne <2g, 8 Airborne <4g */
-#define UBX_TX_CFG_NAV5_DYNMODEL_RTCM	2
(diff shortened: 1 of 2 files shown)


src/ubx.h
```diff
@@ -671,6 +673,9 @@ class GPSDriverUBX : public GPSHelper
 	const Interface		_interface;
 	uint32_t _survey_in_acc_limit;
 	uint32_t _survey_in_min_dur;
+
+	// ublox Dynamic platform model default 7: airborne with <2g acceleration
+	uint8_t _dynModel{7};
```
bkueng: nitpick: naming convention: `_dyn_model`

src/ubx.h
```diff
@@ -571,13 +569,17 @@ class GPSDriverUBX : public GPSHelper
 public:
 	GPSDriverUBX(Interface gpsInterface, GPSCallbackPtr callback, void *callback_user,
 		     struct vehicle_gps_position_s *gps_position,
-		     struct satellite_info_s *satellite_info);
+		     struct satellite_info_s *satellite_info,
+		     uint8_t dynamic_model);
```
bkueng: add default argument `dynamic_model=7`, so that updating the submodule in QGC will just work?

dagar: I actually did that initially, but assumed QGC was the only one setting it to 2?

src/ubx.h
dagar: It was to match the field straight out of ublox documentation, but I'll change it.

src/ubx.h
bkueng: I thought because you have `RTCM mode force stationary dynamic model`, QGC gets it right anyway. I'm ok with either way.

dagar: Right, thanks.

dagar:
Updated

PR #30, opened 2018-08-02
Adding RTCM3 available output message IDs

AlexisTM:
Documentation of the messages on page 370: https://www.u-blox.com/sites/default/files/products/documents/u-blox8-M8_ReceiverDescrProtSpec_%28UBX-13003221%29_Public.pdf

If you ever fly in China, you would need to enable Beidou.

If you are flying with a very restricted data throughput, the MSM4 messages are twice smaller.

If you want to use a moving baseline on a FW; You need the 4072 message.


bkueng:
> If you ever fly in China, you would need to enable Beidou.

I agree we should enable this. Are you in China and can test it?
Future devices will also have Galileo support.

> If you are flying with a very restricted data throughput, the MSM4 messages are twice smaller.

Correct. So far even a 3DR telemetry was enough though.
Generally if you want to extend the interface and add support for it, you are welcome to do that!

PR #35, opened 2018-12-04
PR Support for Septentrio GPS

MatejFranceskin:
Support for Septentrio Binary Format (SBF)

### DIFF
--- a/src/sbf.cpp
+++ b/src/sbf.cpp
@@ -80,101 +80,84 @@ GPSDriverSBF::configure(unsigned &baudrate, OutputMode output_mode)
 	// Check if we're already configured
 	setBaudrate(SBF_TX_CFG_PRT_BAUDRATE);
 
-	if (receive(SBF_CONFIG_TIMEOUT) > 0) {
-		return 0;
+	_output_mode = output_mode;
+	baudrate = 115200;
+
+	setBaudrate(baudrate);
+	// Change the baudrate
+	char msg[64];
+	snprintf(msg, sizeof(msg), SBF_CONFIG_BAUDRATE, baudrate);
+
+	if (!sendMessage(msg)) {
+		return -1; // connection and/or baudrate detection failed
 	}
 
-	unsigned baud_i;
+	if (SBF_TX_CFG_PRT_BAUDRATE != baudrate) {
+		setBaudrate(SBF_TX_CFG_PRT_BAUDRATE);
+		baudrate = SBF_TX_CFG_PRT_BAUDRATE;
+	}
 
-	_output_mode = output_mode;
-	// try different baudrates
-	const unsigned baudrates[] = { 115200 };
+	if (!sendMessageAndWaitForAck(SBF_CONFIG_RESET, SBF_CONFIG_TIMEOUT)) {
+		return -1; // connection and/or baudrate detection failed
+	}
 
-	for (baud_i = 0; baud_i < sizeof(baudrates) / sizeof(baudrates[0]); baud_i++) {
-		baudrate = baudrates[baud_i];
-		SBF_DEBUG("Try baud rate: %d", baudrate);
-		setBaudrate(baudrate);
-		// Change the baudrate
-		char msg[64];
-		snprintf(msg, sizeof(msg), SBF_CONFIG_BAUDRATE, baudrate);
+	// at this point we have correct baudrate on both ends
 
-		if (!sendMessage(msg)) {
-			continue;
-		}
+	const char *config_cmds;
 
-		if (SBF_TX_CFG_PRT_BAUDRATE != baudrate) {
-			setBaudrate(SBF_TX_CFG_PRT_BAUDRATE);
-			baudrate = SBF_TX_CFG_PRT_BAUDRATE;
-		}
+	if (_output_mode == OutputMode::RTCM) {
+		config_cmds = SBF_CONFIG_RTCM;
 
-		if (!sendMessageAndWaitForAck(SBF_CONFIG_RESET, SBF_CONFIG_TIMEOUT)) {
-			continue;
-		}
+	} else {
 
-		// at this point we have correct baudrate on both ends
+		if (_dynamic_model < 6) {
+			snprintf(msg, sizeof(msg), SBF_CONFIG_RECEIVER_DYNAMICS, "low");
 
-		const char *config_cmds;
+		} else if (_dynamic_model < 7) {
+			snprintf(msg, sizeof(msg), SBF_CONFIG_RECEIVER_DYNAMICS, "moderate");
 
-		if (_output_mode == OutputMode::RTCM) {
-			config_cmds = SBF_CONFIG_RTCM;
+		} else if (_dynamic_model < 8) {
+			snprintf(msg, sizeof(msg), SBF_CONFIG_RECEIVER_DYNAMICS, "high");
 
 		} else {
+			snprintf(msg, sizeof(msg), SBF_CONFIG_RECEIVER_DYNAMICS, "max");
+		}
 
-			if (_dynamic_model < 6) {
-				snprintf(msg, sizeof(msg), SBF_CONFIG_RECEIVER_DYNAMICS, "low");
+		sendMessageAndWaitForAck(msg, SBF_CONFIG_TIMEOUT);
 
-			} else if (_dynamic_model < 7) {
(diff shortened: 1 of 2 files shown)


src/sbf.cpp
```diff
@@ -0,0 +1,579 @@
...
+	if (write(msg, length) != length) {
+		return false;
+	}
+
+	// Wait for acknowledge
+	// For all valid set -, get - and exe -commands, the first line of the reply is an exact copy
+	// of the command as entered by the user, preceded with "$R:"
+	char buf[GPS_READ_BUFFER_SIZE];
+	size_t offset = 0;
+	memset(buf, 0, sizeof(buf));
```
bkueng: No need to zero-initialize it, just make sure buf is null-terminated before calling `strstr`.

src/sbf.cpp
```diff
@@ -0,0 +1,579 @@
...
+				// SBF_DEBUG("parsed %d: 0x%x", i, buf[i]);
+			}
+		}
+
+		if (handled > 0) {
+			return handled;
+		}
+
+		// abort after 2s if no useful packets received
+		if (time_started + 2000 * 1000 < gps_absolute_time()) {
```
bkueng: Can you use the `timeout` argument here?

src/sbf.cpp
```diff
@@ -0,0 +1,579 @@
...
+ */
+int    // 0 = no message handled, 1 = message handled, 2 = sat info message handled
+GPSDriverSBF::payloadRxDone()
+{
+	int ret = 0;
+	struct tm timeinfo;
+	time_t epoch;
+	uint8_t *buf_ptr;
+
+	if (_buf.crc16 != crc16(reinterpret_cast<uint8_t *>(&_buf) + 4, _buf.length - 4)) {
```
bkueng: We should check for `_buf.length >= 4` to avoid invalid memory access on corrupt input.

src/sbf.cpp
```diff
@@ -0,0 +1,579 @@
...
+int
+GPSDriverSBF::configure(unsigned &baudrate, OutputMode output_mode)
+{
+	_configured = false;
+
+	// Check if we're already configured
+	setBaudrate(SBF_TX_CFG_PRT_BAUDRATE);
+
+	if (receive(SBF_CONFIG_TIMEOUT) > 0) {
+		return 0;
```
bkueng: Hmm I'd rather reconfigure it even if we receive something. It could still be incorrectly configured in that case.

src/sbf.cpp
```diff
@@ -0,0 +1,579 @@
...
+	}
+
+	unsigned baud_i;
+
+	_output_mode = output_mode;
+	// try different baudrates
+	const unsigned baudrates[] = { 115200, 57600, 38400, 19200, 9600 };
```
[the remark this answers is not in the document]
MatejFranceskin: So far I have seen only 115200 which is also the default. Should we leave only 115200 then and add others later is the use case pops up?

src/sbf.h
```diff
@@ -0,0 +1,384 @@
...
+
+typedef struct {
+	uint16_t sync;              /** The Sync field is a 2-byte array always set to 0x24, 0x40. The first byte of every SBF block has
+									hexadecimal value 24 (decimal 36, ASCII ’$’). The second byte of every SBF block has hexadecimal
+									value 40 (decimal 64, ASCII ’@’). */
+	uint16_t crc16;				/** The CRC field is the 16-bit CRC of all the bytes in an SBF block from and including the ID field
+									to the last byte of the block. The generator polynomial for this CRC is the so-called CRC-CCITT
+									polynomial: x 16 + x 12 + x 5 + x 0 . The CRC is computed in the forward direction using a seed of 0, no
+									reverse and no final XOR. */
+uint16_t msg_id:
```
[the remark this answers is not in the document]
MatejFranceskin: Yes, astyle doesn't seem to handle bit fields correctly.

src/sbf.cpp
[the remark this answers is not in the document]
bkueng: I would say so.

src/sbf.cpp
MatejFranceskin: Implemented.

src/sbf.cpp
MatejFranceskin: Implemented.

src/sbf.cpp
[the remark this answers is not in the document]
MatejFranceskin: Implemented.

src/sbf.cpp
```diff
@@ -0,0 +1,579 @@
...
+
+bool
+GPSDriverSBF::sendMessage(const char *msg)
+{
+	SBF_DEBUG("Send MSG: %s", msg);
+
+	// Send message
+	int length = static_cast<int>(strlen(msg));
+
+	if (write(msg, length) != length) {
```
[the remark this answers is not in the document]
MatejFranceskin: Implemented.

MatejFranceskin:
Module was implemented by copying and then modifying ubx.cpp & ubx.h.

src/sbf.cpp
```diff
@@ -0,0 +1,575 @@
...
+		return 0;
+	}
+
+	unsigned baud_i;
+
+	_output_mode = output_mode;
+	// try different baudrates
+	const unsigned baudrates[] = { 115200 };
+
+	for (baud_i = 0; baud_i < sizeof(baudrates) / sizeof(baudrates[0]); baud_i++) {
```
mhkabir: This loop can be dropped since you're using a fixed baud.

src/sbf.cpp
MatejFranceskin: Yes - original ubx code uses timeout argument. I changed it.

src/sbf.cpp
```diff
@@ -0,0 +1,575 @@
...
+	// For all valid set -, get - and exe -commands, the first line of the reply is an exact copy
+	// of the command as entered by the user, preceded with "$R:"
+	char buf[GPS_READ_BUFFER_SIZE];
+	size_t offset = 0;
+	gps_abstime time_started = gps_absolute_time();
+
+	bool found_response = false;
+
+	do {
+		int ret = read(reinterpret_cast<uint8_t *>(buf) + offset, sizeof(buf) - offset, timeout);
```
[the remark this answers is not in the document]
MatejFranceskin: Thanks - I changed it.

src/sbf.cpp
MatejFranceskin: I agree - I changed it.

src/sbf.cpp
```diff
@@ -0,0 +1,575 @@
...
+
+		if (ret < 0) {
+			// something went wrong when polling or reading
+			SBF_WARN("sbf poll_or_read err");
+			return false;
+		}
+
+		offset += ret;
+		buf[offset] = '\0';
```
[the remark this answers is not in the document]
MatejFranceskin: Commited as suggested. Thanks for pointing it out!

src/sbf.cpp
```diff
@@ -0,0 +1,575 @@
...
+
+		if (!found_response && strstr(buf, "$R: ") != nullptr) {
+			SBF_DEBUG("READ %d: %s", offset, buf);
+			found_response = true;
+		}
+
+		if (offset >= sizeof(buf)) {
+			offset = 0;
+			memset(buf, 0, sizeof(buf));
```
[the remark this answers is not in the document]
MatejFranceskin: Dropped.

src/sbf.cpp
MatejFranceskin: Yes, I dropped it. If we need it in future it is in git history.

hamishwillee:
@bkueng We got a link or any other info for https://docs.px4.io/en/gps_compass/

bkueng:
@hamishwillee you're too quick, we have not flown it yet :)

hamishwillee:
@bkueng YOu have to get up pretty early to be ahead of me (about 10 hours) :-)

Can you make sure it happens at the "right time" (you can just drop me a line if you want). The only trigger I have for stuff like this is PRs, so if you don't then I won't know.

bkueng:
> Can you make sure it happens at the "right time" (you can just drop  me a line if you want). The only trigger I have for stuff like this is  PRs, so if you don't then I won't know.


Yup

PR #38, opened 2019-01-22
gps: Support for Emlid Reach

barchstien:
gps: Support for Emlid Reach
Support Emlid Reach in NMEA and ERB format,
including autodetect

Reported-by: Bastien Auneau <bastien.auneau@while-true.fr>

### DIFF
--- a/src/emlid_reach.cpp
+++ b/src/emlid_reach.cpp
@@ -1,0 +1,377 @@
+/****************************************************************************
+ *
+ *   Copyright (c) 2012-2019 PX4 Development Team. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ * 3. Neither the name PX4 nor the names of its contributors may be
+ *    used to endorse or promote products derived from this software
+ *    without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ * FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ * BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS
+ * OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+ * AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ ****************************************************************************/
+
+/**
+ * @file emlid_reach.cpp
+ *
+ * @author Bastien Auneau <bastien.auneau@while-true.fr>
+ */
+
+
+#include "emlid_reach.h"
+
+#include <stdio.h>
+#include <math.h>
+#include <string.h>
+#include <ctime>
+#include <cmath>
+#include <cstring>
+#include <errno.h>
+
+#include <stdlib.h>
+
+//// ERB
+// 'E' Ox45 | 'R' Ox52 | ID | LENGTH (2 Bytes little endian) | payload | CHECKSUM (2B)
+#define ERB_SYNC_1 0x45 // E
+#define ERB_SYNC_2 0x52 // R
+
+#define ERB_ID_VERSION         0x01
+#define ERB_ID_GEODIC_POSITION 0x02
+#define ERB_ID_NAV_STATUS      0x03
+#define ERB_ID_DOPS            0x04
+#define ERB_ID_VELOCITY_NED    0x05
+#define ERB_ID_SPACE_INFO      0x06 // not used, reduces stack usage
+#define ERB_ID_RTK             0x07 // RTK, TODO
+
+
+#define EMLID_UNUSED(x) (void)x;
+
+#define GPS_PI 3.141592653589793238462643383280f
+
+#define AUTO_DETECT_MAX_READ_SENTENCE 5  // if more detect succeed
+
+
+GPSDriverEmlidReach::GPSDriverEmlidReach(GPSCallbackPtr callback, void *callback_user,
+		struct vehicle_gps_position_s *gps_position, struct satellite_info_s *satellite_info) :
+	GPSHelper(callback, callback_user),
+	_gps_position(gps_position), _satellite_info(satellite_info)
+{}
+
+
+int
+GPSDriverEmlidReach::configure(unsigned &baudrate, OutputMode output_mode)
+{
+	// TODO RTK
+	if (output_mode != OutputMode::GPS) {
+		GPS_WARN("EMLIDREACH: Unsupported Output Mode %i", (int)output_mode);
+		return -1;
+	}
+
+	unsigned baud_allowed[] {57600, 115200, 230400};
+
+	for (unsigned i = 0; i < sizeof(baud_allowed) / sizeof(baud_allowed[0]); i++) {
+		if (baudrate > 0 && baudrate != baud_allowed[i]) {
+			continue;
+		}
+
+		_sentence_cnt = 0;
+		_erb_decode_state  = ERB_State::init;
+		_erb_payload_len = 0;
+		_erb_buff_cnt = 0;
+
+		if (GPSHelper::setBaudrate(baud_allowed[i]) != 0) {
+			continue;
+		}
+
+		if (! testConnection()) {
+			continue;
+		}
+
+		baudrate = baud_allowed[i];
+		return 0;
+	}
+
+	return -1;
+}
+
+
+bool
+GPSDriverEmlidReach::testConnection()
+{
+	_testing_connection = true;
+	receive(500);
+	_testing_connection = false;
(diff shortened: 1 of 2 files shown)


mhkabir:
Wonder if we even want to support the NMEA version since it doesn't give us the full set of data.

src/emlid_reach.cpp
```diff
@@ -0,0 +1,981 @@
+/****************************************************************************
+ *
+ *   Copyright (c) 2012-2015 PX4 Development Team. All rights reserved.
```
dagar: date

src/emlid_reach.h
```diff
@@ -0,0 +1,227 @@
+/****************************************************************************
+ *
+ *   Copyright (c) 2012-2016 PX4 Development Team. All rights reserved.
```
dagar: date

src/emlid_reach.cpp
barchstien: applied change

src/emlid_reach.h
barchstien: change applied

barchstien:
@mhkabir Mohammed, about NMEA support, I have no opinion. I left it there because I started with it and I thought it would be a waste to remove it. But now I understand that it has consequences on the final flash size, which may be a limiting factor (fmu-v2). Do tell me to remove it if you think trade off is leaning that way

mhkabir:
Yeah I think it's generally a good idea to remove it for staying lean on flash. We should only be lugging around utilized code.

Thanks for your contributions!

src/emlid_reach.cpp
```diff
@@ -0,0 +1,434 @@
...
+GPSDriverEmlidReach::testConnection()
+{
+	_testing_connection = true;
+
+	unsigned timeout_cnt = 0;
+
+	while (timeout_cnt < AUTO_DETECT_MAX_TIMEOUT
+	       && _parse_err_cnt < AUTO_DETECT_MAX_PARSE_ERR
+	       && _sentence_cnt < AUTO_DETECT_MAX_READ_SENTENCE) {
+		if (receive(50000) == 0b100) { // timeout larger than what defined in read() are truncated
```
bkueng: Please don't rely on a truncated timeout and use something like 50 or 100 ms here.
I'd also rather check if there is one (or more) successfully parsed message, instead of checking for parser errors.

src/emlid_reach.cpp
```diff
@@ -0,0 +1,434 @@
...
+}
+
+
+int
+GPSDriverEmlidReach::receive(unsigned timeout)
+{
+	while (true) {
+		if (_read_buff_ptr >= _read_buff + _read_buff_len) {
+			// no more data in buff to process, read from serial
+			_read_buff_len = read(_read_buff, sizeof(_read_buff), timeout);
```
bkueng: Instead of reading into `_read_buff` and later on parsing into `_buff`, why not reading into a local buffer `uint8_t buf[GPS_READ_BUFFER_SIZE];` and then directly call `erbParseChar` on that?

src/emlid_reach.cpp
```diff
@@ -0,0 +1,434 @@
...
+
+	if (_buff[2] == ERB_ID_VERSION) {
+		//GPS_INFO("EMLIDREACH: ERB version: %d.%d.%d", _buff[ERB_HEADER_LEN + 4], _buff[ERB_HEADER_LEN + 5], _buff[ERB_HEADER_LEN + 6]);
+
+	} else if (_buff[2] == ERB_ID_GEODIC_POSITION) {
+
+		double tmp_double = 0.0;
+		_gps_position->timestamp = gps_absolute_time();
+
+		memcpy(&_last_POS_timeGPS, _buff + ERB_HEADER_LEN, sizeof(uint32_t));
```
bkueng: If you created a union of (packed) structs for `_buff`, you could avoid the memcpy and offsets, and the code would be more readable.
For example as the u-blox driver does: https://github.com/PX4/GpsDrivers/blob/c8fb76fdeb80a6a579bf0ba303a78f60cc40835c/src/ubx.h#L706.

src/emlid_reach.cpp
```diff
@@ -0,0 +1,434 @@
...
+		struct vehicle_gps_position_s *gps_position, struct satellite_info_s *satellite_info) :
+	GPSHelper(callback, callback_user),
+	_gps_position(gps_position), _satellite_info(satellite_info)
+{}
+
+
+int
+GPSDriverEmlidReach::configure(unsigned &baudrate, OutputMode output_mode)
+{
+	// TODO RTK
```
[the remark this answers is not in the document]
barchstien: Yes, the plan is to work on RTK as soon as the basic GPS has been stabilized and accepted in master. I wanted to be sure I am doing things as expected. About a separate link, nope, we planned to use the very same serial used for getting position

src/emlid_reach.cpp
barchstien: timeout set to 100msec
testConnection() returns true on number of successfully parsed message

src/emlid_reach.cpp
barchstien: GpsHelper::receive() interface returns when there is data to publish. This means that if there is still bytes to be parsed after an ERB (position) sentence is received, those remaining bytes are lost if the buffer is local.
The only way to use a local buff and to be sure there is not data loss, would be that the read() function from the serial device guarantees that data is read line by line. While testing, I could see that data is not always read line by line (even if it appears to reads line by line at first), that is why _read_buff was moved from local scope to class scope

If the aim is to avoid having 2 buffers (_read_buff and _buff), that could be achieved using memmove() and pointer arithmetic

@bkueng what do you say ? Should I go for using 1 buff and memmove()

src/emlid_reach.cpp
barchstien: doing that

src/emlid_reach.cpp
bkueng: The idea would be:
- read into the local buffer
- parse everything that was received for that (via `erbParseChar`)
- if there was a position message parsed, return, otherwise iterate and check for the timeout.

Which means that there are no bytes lost (a partial message would still be kept in `_buff` and continued in the next `receive` call). What could happen is that a `read` contains already multiple position messages. In that case only the last one would be returned, and it means that `receive` is not called fast enough.
Does that happen to you? What is the position update rate?

barchstien: Position update rate is around 5Hz. Practically, it goes between 4.9 a,d 5.1 Hz. When sat view is not good, it may go down (sky is blocked by building for example) if the lock is not stable

I understand the idea, and the limitation associated to it ("only the last would be returned")
I disagree on "it means that receive is not called fast enough", because several message types are received (status, position, ned, etc), they are received by bursts (all types 5 times per sec), and are not all relevant. That means that a message I don't care about (sat info for example), can shadow useful messages (position, ned).
Since ubx is using an rx_buff and a sentence_buff (packed struct), I propose to do the same here (I don't want to change a winning team)
The footprint of the rx_buff is 150, as defined in GpsHelper 
46:#define GPS_READ_BUFFER_SIZE 150

Secondly, regarding the local vs object scope, since the waiting time is happening when serial.read() is called, that means that most of the lifetime of the GPS is happening within the GpsHelper::receive(), which is the local scope mentioned above. I conclude from that that moving the buff to object scope is of very little side effect. Or am I missing something ?

@bkueng Looking forward for your answer :-)

bkueng: >  because several message types are received (status, position, ned, etc)

You would only return on a position update - since that is what we really care about. Then my argument holds (otherwise you're right, it would not).

> I conclude from that that moving the buff to object scope is of very little side effect.

The side effect is the stack size requirement: since other drivers have it on the stack, they impose a minimum stack size to the gps driver. If the emilid driver moves the buffer to the heap (object scope), that stack space is unused, and therefore overall more memory is used.
Or in other words: what's important is that all drivers do (roughtly) the same thing to minimize overall memory usage. Doing it the other way around for all drivers would work too.

barchstien: @bkueng Thanks for the time to answer
Indeed, I didn't consider heap vs stack, just memory vs time... oops :-)
Ok, so I do it on stack, your argument is good, we don't want to waste stack reserved by other drivers

Regarding 
> You would only return on a position update - since that is what we really care about. Then my argument holds (otherwise you're right, it would not).
> 
That is partially true, because some of the info returned with position are obtained with other messages, namely DOPs, NED and fix status.
I think now I'll make the buff var local (so stack), and I will return (the position) only after read() buffer has been fully parsed. That should meet all requirements

barchstien:
@bkueng Commits are rebased and squashed

barchstien:
@bkueng Commits are rebased and squashed

src/emlid_reach.cpp
```diff
@@ -0,0 +1,391 @@
...
+
+	// read from serial
+	read_buff_len = read(read_buff, sizeof(read_buff), timeout);
+
+	if (read_buff_len > 0) {
+		read_ind = 0;
+
+	} else {
+		// timeout occured
+		return 0b100; // no update to publish, but for testConnection() to know about timeout
```
bkueng: This is not conforming with the API definition of `receive()`. We either need to update the API spec and all users of it (QGC and the Firmware), or rather change this to be conforming.

src/emlid_reach.cpp
```diff
@@ -0,0 +1,391 @@
...
+
+
+bool
+GPSDriverEmlidReach::testConnection()
+{
+	_testing_connection = true;
+
+	unsigned timeout_cnt = 0;
+
+	while (timeout_cnt < AUTO_DETECT_MAX_TIMEOUT
```
bkueng: This loop will take 50 * 100 ms = 5 s worst case. Can you reduce that to at most a few 100ms?

src/emlid_reach.cpp
barchstien: @bkueng Yes, indeed. The problem I tried to solve here is that it cannot return 0, or else the loop in gps.cpp exits (line 750)
`while ((helper_ret = _helper->receive(TIMEOUT_5HZ)) > 0 && !should_exit()) {`
I do not have access to should_exit() either
And I didn't want to loop until I have something to receive, or else that could block the gps thread into the driver
What ubx is doing is to exist after 1000 timeout (ubx.cpp line 762)
`/* abort after timeout if no useful packets received */
if (time_started + timeout * 1000 < gps_absolute_time()) {
	UBX_DEBUG("timed out, returning");
	return -1;
}`
I thought that returning a non zero value without the bit set signalling gps or sat update would be more elegant
I also considered changing gps.cpp to
`while ((helper_ret = _helper->receive(TIMEOUT_5HZ)) >= 0 && !should_exit()) {`
But I didn't want to change things outside emlid files

Anyways, I'm changing that to how it is in ubx/ashtech/mtk now

src/emlid_reach.cpp
barchstien: ok

barchstien:
@bkueng code is changed according to your latest comments. I'm actually more satisfied of how it is now, thanks to your feedback :-)
CI worked once with all tests passed this morning (ar 9:30). Then I removed some #define that were not needed anymore, and jenkins tests failed on "no space left on device"... I'll post again when CI has passed all tests

barchstien:
@bkueng CI is back test are all passed

barchstien:
@bkueng Nice :-) 
Should I set the submodule path back to PX4 on https://github.com/PX4/Firmware/pull/11265 ?
It now points to my fork

bkueng:
Yes that should be good to merge then as well.

barchstien:
> Yes that should be good to merge then as well.

Done :-)

Issue #26, opened 2018-02-02
Add UBX-CFG-RST to force a Cold Start


AlexKlimaj:
I am finding that the Ublox MAX-M8 can get in a bad state when being powered on and off indoors for short periods of time. Once taken outside, it can take 1-2 hours to acquire satellites. 

When booting up a fresh MAX-M8 outside, it only takes ~45 seconds to acquire satellites. 

On a unit that got into a bad state, I connected it to Ublox U-center and forced a Cold start. The unit then acquired satellites in ~45 seconds. On this design, I am averaging 18-22 satellites, so I'm certain the RF frontend is working well. 

The UBX protocol on u-blox 8/M8 from protocol 15 up supports a Cold boot command using UBX-CFG-RST defined in section 31.11.26 UBX-CFG-RST (0x06 0x04) of [u-blox8-M8_ReceiverDescrProtSpec_(UBX-13003221)](https://www.u-blox.com/sites/default/files/products/documents/u-blox8-M8_ReceiverDescrProtSpec_%28UBX-13003221%29.pdf)

dagar:
How would you see this being used? Can we detect the bad state?

AlexKlimaj:
I see it being accessible through the nut shell gps app, mavlink, and a qgroundcontrol button. 

I think a bad state could automatically be detected in PX4 if a GPS module is connected and the satellite count has been equal to 0 for X minutes.

dagar:
@davids5 fyi

davids5:
@bkueng - Relative to having to test for interference, and being able establish a baseline.  
I am wondering if you know if these devices can be told to use warm or cold start?

AlexKlimaj:
In the Ublox U-center, there are buttons to initiate a warm start, hot start, or cold start.

AlexKlimaj:
From 
https://www.u-blox.com/sites/default/files/products/documents/u-blox6_ReceiverDescrProtSpec_%28GPS.G6-SW-10018%29_Public.pdf 
on page 17:

> In the 
UBX-CFG-RST
 message, one can force the receiver to reset and clear data, in order to see the effects of
maintaining/losing such data between restarts. For this, the CFG-RST message offers the 
navBbrMask
 field,
where Hot, Warm and Cold starts can be initiated, and also other combinations thereof.
The Reset Type can also be specified. This is not related to GPS, but to the way the software restarts the
system.

davids5:
[...]

PR #40, opened 2019-02-06
PR GPS Reset

MatejFranceskin:
Solution for issue #26 Add UBX-CFG-RST to force a Cold Start

### DIFF
--- a/src/gps_helper.h
+++ b/src/gps_helper.h
@@ -98,6 +98,13 @@ enum class GPSCallbackType {
 	setClock,
 };
 
+enum class GPSRestartType {
+	None,
+	Hot,
+	Warm,
+	Cold
+};
+
 /** Callback function for platform-specific stuff.
  * data1 and data2 depend on type and user is the custom user-supplied argument.
  * @return <0 on error, >=0 on success (depending on type)
@@ -150,6 +157,15 @@ public:
 	 *         bit 1 set: got satellite info update
 	 */
 	virtual int receive(unsigned timeout) = 0;
+
+	/**
+	 * Reset GPS device
+	 * @param restart_type
+	 * @return -1 if not implemented
+	 * 	        0 on failure
+	 *          1 on success
+	 */
+	virtual int reset(GPSRestartType restart_type)	{ return -1; }
 
 	float getPositionUpdateRate() { return _rate_lat_lon; }
 	float getVelocityUpdateRate() { return _rate_vel; }
--- a/src/gps_helper.h
+++ b/src/gps_helper.h
@@ -101,19 +101,28 @@ enum class GPSCallbackType {
 enum class GPSRestartType {
 	None,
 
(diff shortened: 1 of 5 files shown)


src/gps_helper.h
```diff
@@ -151,6 +158,15 @@ class GPSHelper
 	 */
 	virtual int receive(unsigned timeout) = 0;
 
+	/**
+	 * Reset GPS device
+	 * @param restart_type
+	 * @return -1 if not implemented
+	 * 	        0 on failure
```
bkueng: Can you return 0 on success, and <0 for failures?

src/gps_helper.h
```diff
@@ -98,6 +98,13 @@ enum class GPSCallbackType {
 	setClock,
 };
 
+enum class GPSRestartType {
+	None,
+	Hot,
```
bkueng: Can you add some documentation what the exact meaning is for each type?

src/ubx.h
```diff
@@ -202,6 +204,15 @@
 #define UBX_TX_CFG_NAV5_MASK		0x0005		/**< Only update dynamic model and fix mode */
 #define UBX_TX_CFG_NAV5_FIXMODE		2		/**< 1 2D only, 2 3D only, 3 Auto 2D/3D */
 
+/* TX CFG-RST message contents
+ * Note: not used with protocol version 27+ anymore
```
bkueng: Can you add a TODO in the code or use the new message if `_proto_ver_27_or_higher` is set?

src/gps_helper.h
MatejFranceskin: Added descriptions from UBX documentation.

src/gps_helper.h
MatejFranceskin: Some GPS drivers do not have reset function, so I changed it like this:
0 : success
-1: not implemented
<0: failure

src/ubx.h
MatejFranceskin: Removed the comment - it was a copy/paste error and it was not vaild for CFG-RST

MatejFranceskin:
Thank you for the review - I improved the code and rebased so that everything is in sync with upstream.

src/gps_helper.h
```diff
@@ -98,6 +98,33 @@ enum class GPSCallbackType {
...
+	 * In hot start mode, the receiver was powered down only for a short time (4 hours or less),
+	 * so that its ephemeris is still valid. Since the receiver doesn't need to download ephemeris
+	 * again, this is the fastest startup method.
+	 */
+	Hot,
+
+	/**
+	 * In warm start mode, the receiver has approximate information for time, position, and coarse
+	 * satellite position data (Almanac). In this mode, after power-up, the receiver normally needs
+	 * to downloadephemeris before it can calculate position and velocity data.
```
bkueng: typo: `downloadephemeris`

PR #45, opened 2019-06-20
【WIP】 ubx【RTK】:  Add event info for timestamp.

jiejiezhang:
When we perform 3D modeling of surveying tasks, in order to prevent the delay of the serial port, we need to get the precise time of taking photos.

This is from ubx data sheet：
picture

### DIFF
--- a/src/ubx.cpp
+++ b/src/ubx.cpp
@@ -83,10 +83,12 @@
 GPSDriverUBX::GPSDriverUBX(Interface gpsInterface, GPSCallbackPtr callback, void *callback_user,
 			   struct vehicle_gps_position_s *gps_position,
 			   struct satellite_info_s *satellite_info,
+			   struct event_info_s *event_info,
 			   uint8_t dynamic_model)
 	: GPSBaseStationSupport(callback, callback_user)
 	, _gps_position(gps_position)
 	, _satellite_info(satellite_info)
+	, _event_info(event_info)
 	, _interface(gpsInterface)
 	, _dyn_model(dynamic_model)
 {
@@ -417,6 +419,10 @@ int GPSDriverUBX::configureDevicePreV27()
 		if (!configureMessageRateAndAck(UBX_MSG_NAV_VELNED, 1, true)) {
 			return -1;
 		}
+	}
+
+	if (!configureMessageRateAndAck(UBX_MSG_TIM_TIM2, 5, true)) {
+		return -1;
 	}
 
 	if (!configureMessageRateAndAck(UBX_MSG_NAV_DOP, 1, true)) {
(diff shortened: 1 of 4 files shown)


src/ubx.h
```diff
@@ -62,10 +62,12 @@
 #define UBX_CLASS_ACK		0x05
 #define UBX_CLASS_CFG		0x06
 #define UBX_CLASS_MON		0x0A
-#define UBX_CLASS_RTCM3	0xF5
+#define UBX_CLASS_EVENT     0x0D
+#define UBX_CLASS_RTCM3	0xF5 /**< This is undocumented (?) */
 
 /* Message IDs */
 #define UBX_ID_NAV_POSLLH	0x02
+#define UBX_ID_NAV_EVENT    0X03
```
xdwgood: the format is not uniform,or just a display issue on github web?

src/ubx.cpp
```diff
@@ -1664,6 +1666,23 @@ GPSDriverUBX::payloadRxDone()
 		ret = 1;
 		break;
 
+	case UBX_MSG_TIM_TIM2:
+		UBX_TRACE_RXMSG("Rx TIM-TIM2");
+
+		{
+			// convert to unix timestamp
+			uint64_t tv_sec = _buf.payload_rx_tim_tim2.wnR * 7 * 24 * 3600;
```
xdwgood: Just note: If we need add 1980.01.06(gps start time)?

src/ubx.cpp
```diff
@@ -1664,6 +1666,23 @@ GPSDriverUBX::payloadRxDone()
...
+
+			_event_info->time_utc_usec = static_cast<uint64_t>(tv_sec) * 1e6 +
+						static_cast<uint64_t>_buf.payload_rx_tim_tim2.towMsR * 1e3 +
+						_buf.payload_rx_tim_tim2.towSubMsF / 1e3;
+			_event_info->timestamp = gps_absolute_time();
+
+		}
+
+		ret = 1;
```
xdwgood: still return 1?

bkueng: We'll need to add another bit, so return 4.

src/ubx.h
jiejiezhang: Just a display error on the web page

src/ubx.cpp
jiejiezhang: Thanks. It have been fixed.

src/ubx.cpp
jiejiezhang: Now just based on the data sheet, I can't confirm if it need to add.

jiejiezhang:
FYI: @lumeier

src/ubx.cpp
bkueng: Can you check that? According to the datasheet it depends on the `flags` bits.

bkueng:
Is there any setup configuration required to use it? What is the setup you use?

We have a rising and falling edge timestamp. Can we add both?

There are some `flags` bits that are important as well. For instance we need to check it the `time` is valid, and whether we have utc time.
Then we should also add the `newRisingEdge` information to the message.

jiejiezhang:
It seems no need,

bkueng:
Why, did you end up using a different solution?

PR #46, opened 2019-10-28
clang-tidy: fix hicpp-member-init

dagar:


src/ashtech.h
```diff
@@ -121,7 +121,7 @@ class GPSDriverAshtech : public GPSBaseStationSupport
 	uint64_t _last_timestamp_time{0};
 
 	NMEADecodeState _decode_state{NMEADecodeState::uninit};
-	uint8_t _rx_buffer[ASHTECH_RECV_BUFFER_SIZE];
+	uint8_t _rx_buffer[ASHTECH_RECV_BUFFER_SIZE] {};
```
bkueng: This is one of the places I'd rather keep that way - It's just wasting cycles.

src/ubx.cpp
```diff
@@ -1394,7 +1394,7 @@ GPSDriverUBX::payloadRxDone()
 		    && (_buf.payload_rx_nav_pvt.valid & UBX_RX_NAV_PVT_VALID_VALIDTIME)
 		    && (_buf.payload_rx_nav_pvt.valid & UBX_RX_NAV_PVT_VALID_FULLYRESOLVED)) {
 			/* convert to unix timestamp */
-			struct tm timeinfo;
+			struct tm timeinfo {};
 			memset(&timeinfo, 0, sizeof(timeinfo));
```
bkueng: Please remove this.

src/ashtech.h
dagar: Yes that's fine, this is just an experiment to see if we could have this enforced everywhere with a tool at the cost of occasional wasted effort. The current orb patterns are the main thing in the way that would actually be cost prohibitive.

DonLakeFlyer:
Anything here need to come to QGC?

bkueng:
No, this does not fix anything.

PR #47, opened 2019-11-06
[CRITICAL] SBF fix for do-not-use values

AndreasAntener:
We have encountered a case on an AsteRx-m2 UAS where do-not-use values were being reported in velocity fields without a corresponding error code or fix-type 0. Support confirmed that this is happening. This adds a check for known invalid data and sets the appropriate flags.

Unfortunately I have currently no hardware to test this anymore. Not 100% sure if this is the best way to do the checks.

Note that if the glitches are short, EKF will reject the extreme velocity values and it goes unnoticed. Other estimators might behave differently.

do-not-use value in this case was `-2 * 10^10`

### DIFF
--- a/src/sbf.cpp
+++ b/src/sbf.cpp
@@ -415,7 +415,22 @@ GPSDriverSBF::payloadRxDone()
 			_gps_position->fix_type = 3;
 		}
 
+		// Check fix and error code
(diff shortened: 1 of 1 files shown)


AndreasAntener:
Note that the m2 reference specifies in a very broad and general sense that these do-not-use values might occur at anytime in any field (in theory). We might want to extend the checks further.

PR #51, opened 2020-03-16
ubx: fix baud rate search

NicolasM0:
Bug:
In Auto baud rate, if the gps current baud rate is not the first tested,
desired baud rate will be set to the pre 2.7 default bauderate
(UBX_TX_CFG_PRT_BAUDRATE)
For 2.7 and high it should be set to UBX_CFG_KEY_CFG_UART1_BAUDRATE

Solution:
Wait to be sure that protocol version is lower than 2.7 before setting
desired_baudrate to UBX_TX_CFG_PRT_BAUDRATE


NicolasM0:
I tested on an F9P

PR #54, opened 2020-04-05
fixed unused arg compiler warning

dakejahl:
This throws a warning, in QGC warnings are errors so QGC master currently does not compile.


LorenzMeier:
Thanks!

PR #55, opened 2020-04-09
ubx: support NEO-M9N

bkueng:
RTK configuration is not supported, thus:
- only configure RTCM output for the base station
- make UBX_CFG_KEY_NAVHPG_DGNSSMODE optional

Fixes https://github.com/PX4/GpsDrivers/issues/50

Tested on F9P + M9N


LorenzMeier:
Could you please assign a reviewer going forward? Otherwise people don't get notified.

bkueng:
And who should that be?
Obviously you did see it.

LorenzMeier:
Yes, but because I'm the "fixer". If I disable GH notifications for everything PX4 will fall apart because a lot of these things will fall through the cracks. But that doesn't work any more even for me because PX4 is growing. I prefer if you directly tag me than nobody - I can still hand it off.

Thanks!

PR #57, opened 2020-04-24
u-blox: reduce update rate for M9N to 5Hz (instead of 10Hz)

bkueng:
Apparently a 10Hz update rate restricts the number of used satellites to 16, and as soon as the update rate is reduced, the number of used satellites increases.

The datasheet only mentions that the navigation rate can go up to 25Hz.

Cross-tested on an F9P.

### DIFF
--- a/src/ubx.cpp
+++ b/src/ubx.cpp
@@ -455,7 +455,10 @@ int GPSDriverUBX::configureDevice()
 	cfgValset<uint8_t>(UBX_CFG_KEY_ODO_OUTLPCOG, 0, cfg_valset_msg_size);
 
 	// measurement rate
-	cfgValset<uint16_t>(UBX_CFG_KEY_RATE_MEAS, 100 /* 10 Hz update rate */, cfg_valset_msg_size);
+	// In case of F9P we use 10Hz, otherwise 5Hz (receivers such as M9N can go higher as well, but
+	// the number of used satellites will be restricted to 16. Not mentioned in datasheet)
+	const int rate_meas = (_board == Board::u_blox9_F9P) ? 100 : 200;
+	cfgValset<uint16_t>(UBX_CFG_KEY_RATE_MEAS, rate_meas, cfg_valset_msg_size);
 	cfgValset<uint16_t>(UBX_CFG_KEY_RATE_NAV, 1, cfg_valset_msg_size);
 	cfgValset<uint8_t>(UBX_CFG_KEY_RATE_TIMEREF, 0, cfg_valset_msg_size);
 
@@ -1330,6 +1333,15 @@ GPSDriverUBX::payloadRxAddMonVer(const uint8_t b)
 		if (buf_index == sizeof(ubx_payload_rx_mon_ver_part2_t) - 1) {
 			// Part 2 complete: decode Part 2 buffer
 			UBX_DEBUG("VER ext \" %30s\"", _buf.payload_rx_mon_ver_part2.extension);
+
+			// in case of u-blox9 family, check if it's an F9P
+			if (_board == Board::u_blox9) {
+				if (strstr((const char *)_buf.payload_rx_mon_ver_part2.extension, "MOD=") &&
+				    strstr((const char *)_buf.payload_rx_mon_ver_part2.extension, "F9P")) {
+					_board = Board::u_blox9_F9P;
+					UBX_DEBUG("F9P detected");
+				}
+			}
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


julianoes:
@bkueng and this has nothing to do with baudrate or number of messages configured?

ryanjAA:
That’s too bad about the 16 sat limitation. 

This commit breaks functionality on the m9n reverted to previous submodule fix and changed hz to 5 and another to 20. 

I have one unit sitting at 5hz and one at 20hz (for good measure). Will report back on max sats.

ryanjAA:
16 sats on 25hz (changed it to max to try) and >20 sats at 5hz so it’s strange that at anything 10hz and above it drops it to 16 sats no matter what but still needed to use your previous driver change for it to work (with amendments to the ubx.cpp file to test sat vs hz).

bkueng:
> @bkueng and this has nothing to do with baudrate or number of messages configured?

No, I played around in u-center with different settings.

> That’s too bad about the 16 sat limitation.

Yes that's unfotunate. It makes sense though as the ZED-F9P has the larger form factor and thus more compute.

ryanjAA:
This breaks m9n functionality though completely.

bkueng:
How?

ryanjAA:
It never finds the device. Just keeps cycling through the GPS probe at different bauds. Tried also changing to 115200 and still no go but using the exact hardware and just flashing a build with the previous driver change and it finds it immediately (was wanting to remove this exact piece of hardware as a variable). 

Screen Shot 2020-04-28 at 11 06 00 AM
Screen Shot 2020-04-28 at 11 05 45 AM

bkueng:
Which exact gps device is that?

And just that we use the same code, can you test with https://github.com/PX4/Firmware/pull/14785?

And to further debug can you enable debug output in https://github.com/PX4/GpsDrivers/blob/master/src/ubx.cpp#L80, and either start the gps driver from the shell or use `dmesg` to get the output?

ryanjAA:
Devices were made in house, the M8N modules work fine and these work fine with the other PR changes, just not this one.

Tried with https://github.com/PX4/Firmware/pull/14785

GPS Debug enabled. It is timing out after receiving UBX msg 0x068a (see attached).


Screen Shot 2020-04-29 at 11 58 30 AM

bkueng:
Thanks. So it's timing out on the first UBX_MSG_CFG_VALSET. And you say reverting cdee27f17e4df7da50090edd6a0243341f6b25b6 makes it work?
Just want to double-check, because that would be really strange, as cdee27f17e4df7da50090edd6a0243341f6b25b6 only affects code happening after the first UBX_MSG_CFG_VALSET.

ryanjAA:
No problem. No, https://github.com/PX4/GpsDrivers/commit/cdee27f17e4df7da50090edd6a0243341f6b25b6 doesn’t work. It’s the initial m9n addition that works, https://github.com/PX4/GpsDrivers/commit/4cc28b8a4dfa8962a399b0079e5da28f1fabd2f6 That as we know has the 16 sat limit so then changing the update rate in the cpp file from 100 to 200 manually (10hz to 5hz) in https://github.com/PX4/GpsDrivers/commit/4cc28b8a4dfa8962a399b0079e5da28f1fabd2f6 has full functionality and unrestricted sat usage.

ryanjAA:
Whats the difference between this and master? If i flash master as of right now from QGC, it works. If i pull https://github.com/PX4/GpsDrivers/commit/cdee27f17e4df7da50090edd6a0243341f6b25b6 in as i did from ~2 days ago, it does not. Normally is wouldn't matter but if this is trying to be put into 11 release this needs to work with that and not master.

cturvey:
The ceiling looks to be 8 Hz (125ms), not 5 Hz
FW SPG 4.03 using 28-30 satellites at 8 Hz
-Clive

https://portal.u-blox.com/s/question/0D52p0000AOK91vCQD/can-neom9n-only-use-16-satellites-with-nav-update-rate-5hz

bkueng:
Thanks @cturvey. Did you test that already with this driver? If so we can update it to 8 Hz.

cturvey:
This is a receiver level observation, I'm not using this driver.
The threshold condition to 16 satellites used is 100 ms or lower rate. It continues to track a fuller complement of satellites, it just doesn't use these in the solution.
Testing indicated 101 ms (9.9 Hz) was also workable for higher satellite counts, but 125 ms (8 Hz) and 200 ms (5 Hz) are definitely usabled. The benefit of 125 ms being it is still divisible into 1 second.
I can't speak to the knock-on effect of these speeds with other aspects of the drivers.
I will push the issue into uBlox to document better (ceiling, choice of satellites, etc), or move the threshold.
-Clive

bkueng:
Ok, I'll do the testing then and adjust accordingly

PR #58, opened 2020-04-24
Add NMEA-0183 Protocol GPS Driver

bys1123:
Support standard NMEA messages. 
Such as support more than 15 satellites which ashtech driver restricted in GGA message, and GSA also not working well.

Not implement baud rate detection yet. 

Due to `sensor_gps_s` changes, not tested with current master.

### DIFF
--- a/src/nmea.cpp
+++ b/src/nmea.cpp
@@ -1,0 +1,866 @@
+/****************************************************************************
+ *
+ *   Copyright (c) 2012-2020 PX4 Development Team. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ * 3. Neither the name PX4 nor the names of its contributors may be
+ *    used to endorse or promote products derived from this software
+ *    without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ * FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ * BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS
+ * OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+ * AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ ****************************************************************************/
+
+/** 
+ * @file nmea.cpp
+ * 
+ * NMEA protocol implementation. 
+ * 
+ * @author WeiPeng Guo <guoweipeng1990@sina.com>
+ * @author Stone White <stone@thone.io>
+ * 
+ */
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <math.h>
+#include <string.h>
+#include <ctime>
+
+#include "nmea.h"
+#include <systemlib/mavlink_log.h>
+
+extern orb_advert_t mavlink_log_pub;
+
+GPSDriverNMEA::GPSDriverNMEA(GPSCallbackPtr callback, void *callback_user,
+				   sensor_gps_s *gps_position,
+				   satellite_info_s *satellite_info,
+				   uint32_t nmea_baud)
+	: GPSHelper(callback, callback_user)
+	, _satellite_info(satellite_info),
+	, _gps_position(gps_position),
+	, _baudrate(nmea_baud)
+{
+	decodeInit();
+	_decode_state = NMEA_DECODE_UNINIT;
+	_rx_buffer_bytes = 0;
+}
+
+/*
+ * All NMEA descriptions are taken from
+ * http://www.trimble.com/OEM_ReceiverHelp/V4.44/en/NMEA-0183messages_MessageOverview.html
+ */
+
+int GPSDriverNMEA::handleMessage(int len)
+{
+	char *endp;
+
+	if (len < 7) {
+		return 0;
+	}
+
+	int uiCalcComma = 0;
+
+	for (int i = 0 ; i < len; i++) {
+		if (_rx_buffer[i] == ',') { uiCalcComma++; }
+	}
+
+	char *bufptr = (char *)(_rx_buffer + 6);
+	int ret = 0;
+
+    	if ((memcmp(_rx_buffer + 3, "ZDA,", 3) == 0) && (uiCalcComma == 6)) {
+		/*
+		UTC day, month, and year, and local time zone offset
+		An example of the ZDA message string is:
+
+		$GPZDA,172809.456,12,07,1996,00,00*45
+
+		ZDA message fields
+		Field	Meaning
+		0	Message ID $GPZDA
+		1	UTC
+		2	Day, ranging between 01 and 31
+		3	Month, ranging between 01 and 12
+		4	Year
+		5	Local time zone offset from GMT, ranging from 00 through 13 hours
+		6	Local time zone offset from GMT, ranging from 00 through 59 minutes
+		7	The checksum data, always begins with *
+		Fields 5 and 6 together yield the total offset. For example, if field 5 is -5 and field 6 is +15, local time is 5 hours and 15 minutes earlier than GMT.
+		*/
+		double nmea_time = 0.0;
+		int day = 0, month = 0, year = 0, local_time_off_hour __attribute__((unused)) = 0,
+		    local_time_off_min __attribute__((unused)) = 0;
+
+		if (bufptr && *(++bufptr) != ',') { nmea_time = strtod(bufptr, &endp); bufptr = endp; }
+
+		if (bufptr && *(++bufptr) != ',') { day = strtol(bufptr, &endp, 10); bufptr = endp; }
+
+		if (bufptr && *(++bufptr) != ',') { month = strtol(bufptr, &endp, 10); bufptr = endp; }
+
+		if (bufptr && *(++bufptr) != ',') { year = strtol(bufptr, &endp, 10); bufptr = endp; }
+
+		if (bufptr && *(++bufptr) != ',') { local_time_off_hour = strtol(bufptr, &endp, 10); bufptr = endp; }
+
+		if (bufptr && *(++bufptr) != ',') { local_time_off_min = strtol(bufptr, &endp, 10); bufptr = endp; }
+
+
+		int nmea_hour = static_cast<int>(nmea_time / 10000);
+		int nmea_minute = static_cast<int>((nmea_time - nmea_hour * 10000) / 100);
+		double nmea_sec = static_cast<double>(nmea_time - nmea_hour * 10000 - nmea_minute * 100);
+
+		/*
+		 * convert to unix timestamp
+		 */
+		struct tm timeinfo = {};
+		timeinfo.tm_year = year - 1900;
+		timeinfo.tm_mon = month - 1;
+		timeinfo.tm_mday = day;
+		timeinfo.tm_hour = nmea_hour;
+		timeinfo.tm_min = nmea_minute;
+		timeinfo.tm_sec = int(nmea_sec);
+		timeinfo.tm_isdst = 0;
+
+#ifndef NO_MKTIME
+		time_t epoch = mktime(&timeinfo);
+
+		if (epoch > GPS_EPOCH_SECS) {
+			uint64_t usecs = static_cast<uint64_t>((nmea_sec - static_cast<uint64_t>(nmea_sec))) * 1000000;
+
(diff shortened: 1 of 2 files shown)


bkueng:
Thanks @bys1123 

> Such as support more than 15 satellites which ashtech driver restricted.

Where do you see that limitation?
On which device was this tested?

bys1123:
@bkueng On M8N, you could see GGA message description in u-center.
And here is a community information.
https://portal.u-blox.com/s/question/0D52p00008RmGUMCA3/number-of-satellites-in-gga

bys1123:
image
Test console output.

bys1123:
image
Test only connect GPS2.

bkueng:
We use the ubx driver for u-blox devices, so I don't see the use for this.

bys1123:
Yes, never use NMEA on U-Blox GPS, I just say you could test with it, I know you must at least have one.

There is a batch of GPS brand is using NMEA protocol as it's a industrial standard, and their performance is enough for drones, some models even better than U-blox.

bys1123:
And they all restricted by GGA message.

bys1123:
Like NovAtel:
image

bkueng:
> And they all restricted by GGA message.

Ok, that is helpful to know. Can we extend the ashtech driver to support these? What difference is required?

bys1123:
image
SinoGNSS

bys1123:
1. Use ashtech name for general NMEA protocol is misleading.

bkueng:
Ok that is fine. A few points:
- check if code sharing between the 2 drivers makes sense to reduce flash size (ashtech driver is 7KB)
- provide an exact list of tested and supported devices (also if that includes RTK)
- auto-detection: how do we make sure the nmea driver does not run on u-blox devices (for instance)? At this point we might have to abandon auto-detection and add a protocol parameter (as you already started with) and default to u-blox.

lukegluke:
I do something similar for my own and it is not universal, I can't share it. But I have some thoughts for you, hope they would be useful.

1) To be truly universal, nmea driver should somehow autodetect what sentence is last in set to return 1 (it could easy be RMC or VTG for example) . For now it can be only GGA in suggested driver.
The most difficult thing here I think will be not be broken by possible rare sentences. 
2) ["HDT" sentence](https://www.trimble.com/OEM_ReceiverHelp/V4.44/en/NMEA-0183messages_HDT.html) should be also parsed in driver it is useful for two antenna RTK receivers with heading. Ashtech driver parse it already.
3) I would recommend to use uiCalcComma >= 14 when checking "GGA", because I encountered the receiver with extra "," in the end. Maybe this is just particular receiver bug, but it is not critical to check only min number of fields/commas going to parse in driver.
4) Missed checking number of commas when checking "RMC". Should be uiCalcComma >= 11
5) GPRMC message fields meaning comment is incorrect. It is copied from the trimble [site](https://www.trimble.com/OEM_ReceiverHelp/V4.44/en/NMEA-0183messages_RMC.html) I suppose? It omits "N or S", "E or W" indicators.
6) Mistake
`uint64_t usecs = static_cast<uint64_t>((nmea_sec - static_cast<uint64_t>(nmea_sec))) * 1000000;`
should be
`uint64_t usecs = static_cast<uint64_t>((nmea_sec - static_cast<uint64_t>(nmea_sec)) * 1000000);`
to fix missing parsed micro seconds. **By the way this bug is from current ashtech driver**.
7) I would recommend to fix parsing like this (here _bytes_readed is my name for bytes_count): 
```
        /* pass received bytes to the packet decoder */
        while (_bytes_parsed < _bytes_readed) {
            int l = 0;

            if ((l = parseChar(_read_buf[_bytes_parsed++])) > 0) {
                /* return to configure during configuration or to the gps driver during normal work
                 * if a packet has arrived */
		int ret = handleMessage(l);

		if (ret > 0) {
			return ret;
	        }
            }
        }

        /* everything is read */
        _bytes_parsed = _bytes_readed = 0;
```
where _bytes_parsed is GPSDriverNmea member variable and is cleared in the same places as _bytes_readed. Or you will miss data on handling satellite info message in future. Take care to increase _bytes_parsed immediately to avoid sending same byte to parcer.
**For current time this is more related to current ashtech driver** I think.
8) I promote suggestion to include ashtech withing nmea

bys1123:
@lukegluke Thanks for review and these nice suggestions, I'll take research on those.

bys1123:
1. Automatic match the baud rate after detect NMEA-0183 protocol.
2. Add GPHDT message support.
3. Support satellite num display for modules only provide “GBGSV, BDGSV, GLGSV” message.
4. Optimized GGA message logic.
5. Reset FC clock after received ZDA or RMC message.
6. Optimized comma handling.

src/nmea.cpp
```diff
@@ -0,0 +1,1048 @@
...
+		int nmea_mth = static_cast<int>((nmea_date - nema_day * 10000) / 100);
+		int nmea_year = static_cast<int>(nmea_date - nema_day * 10000 - nmea_mth * 100);
+		/* convert from degrees, minutes and seconds to degrees * 1e7 */
+		_gps_position->lat = static_cast<int>((int(lat * 0.01) + (lat * 0.01 - int(lat * 0.01)) * 100.0 / 60.0) * 10000000);
+		_gps_position->lon = static_cast<int>((int(lon * 0.01) + (lon * 0.01 - int(lon * 0.01)) * 100.0 / 60.0) * 10000000);
+
+		_gps_position->vel_m_s = velocity_ms;
+		_gps_position->vel_n_m_s =velocity_north;
+		_gps_position->vel_e_m_s =velocity_east;
+		_gps_position->cog_rad =track_rad;
```
lukegluke, line 488: _gps_position->cog_rad = wrap_pi(track_rad);

because track_rad is [0, 2π), while cog_rad  needs to be [-π, π]

src/nmea.cpp
```diff
@@ -0,0 +1,1048 @@
...
+
+		float track_rad = static_cast<float>(track_true) * NMEA_PI/ 180.0f;
+		float velocity_ms =static_cast<float>(ground_speed) / 1.9438445f;
+		float velocity_north =static_cast<float>(velocity_ms) * cosf(track_rad);
+		float velocity_east  = static_cast<float>(velocity_ms) *sinf(track_rad);
+
+		_gps_position->vel_m_s = velocity_ms;
+		_gps_position->vel_n_m_s =velocity_north;
+		_gps_position->vel_e_m_s =velocity_east;
+		_gps_position->cog_rad =track_rad;
```
lukegluke, line 818: same here

moreba1:
https://github.com/ArduPilot/ardupilot/blob/997ffcf30d636c2236c21bc1bb1448643a938c2a/libraries/AP_GPS/AP_GPS_NMEA.cpp

does this help? 

@lukegluke 
@bkueng

lukegluke:
> does this help?

There no problems with current pull request, it is adopted for PX4 and works correctly, while the code you linked at is quite separate thing and also under GPL license. I don't get it why you mentioned it?

moreba1:
So why not get merged?

moreba1:
@bys1123 

can you correct this driver?

bys1123:
@lukegluke I'm very busy on my work these days, I invited you to my project, can you help optimize the code, thanks!

lukegluke:
@bys1123 Thanks, but unfortunately I'm also very busy :(, but at least on next week I will add those modifications I mentioned previously in comments.

YANG-DEMIN:
> Support standard NMEA messages.
> Such as support more than 15 satellites which ashtech driver restricted in GGA message, and GSA also not working well.
> 
> Not implement baud rate detection yet.
> 
> Due to `sensor_gps_s` changes, not tested with current master.

Hello, stone.
I am adjusting your code to the latest release. As you said, the `sensor_gps_s` have changed to `vehicle_gps_position`. In the compile process,  this code `_gps_position->nmea_protol_detected = true; ` meet the error, because the `vehicle_gps_position` struct does not have the `nema_protol_detected` number. 
So I want to know,  how do you fit the nema.cpp to your code.
Think you very much!

lukegluke:
I've pushed commit with fixes I mentioned and also several optimizations and fix code style. Also I tried to add same modifications as it was done in latest ashtech.

But I don't think this should be merged as separate driver.
Instead it should be `GPSBaseStationSupport` class and combined with ashtech.

I see this as general nmea driver with probably additional subsystem flag (ashtech or other receivers for future use). Also it occupy quite a space and needs more optimizations.
Sorry I don't have time for this, especially don't have opportunity and px4 hardware for tests.

siddux:
Hi, I was trying the NMEA protocol implementation mentioned above on a Pixhawk cube black using a Swift Nav receiver. I can get coordinates info on the nutshell but the gps status still remains "No ok". Any idea on how to fix it? 

gps_status

Thanks.

lukegluke:
VTG has a bug, I forgot to fix it here. It should be fixed to 8 commas:
`} else if ((memcmp(_rx_buffer + 3, "VTG,", 4) == 0) && (uiCalcComma >= 8)) {`

@siddux try this

jajberni:
I have been trying to test this pull request but unfortunately, we found some issues because of the new format in the constructor and the fact that messages don't necessarily have the same rates. The current implementation waits until all the messages arrive before returning OK, which causes the `NOT OK` status. In the Ashtech implementation, every time that a valid position or speed is received, it returns 1. I testes this approach and now we get 10Hz rates for speed and velocity. Satellite information and heading update the `_gps_position` as they arrive. 

The implementation, based on DimianZhan:add_nmea_driver is at: 

https://github.com/OpenAgriTech/PX4-GPSDrivers/tree/nmea_support. 

However, due to the changes in the master branch, I had to fork it from the master and I am not sure how to reflect my changes from DimianZhan:add_nmea_driver. 

This seems to work well and we have now RTK and heading without any issues using an RTKite with a dual antenna.

bkueng:
@jajberni can you open a pull request?

lukegluke:
> I have been trying to test this pull request but unfortunately, we found some issues because of the new format in the constructor

I suppose you meant the configure() function, not the constructor. Yes, there is a minor change GPSConfig.

> and the fact that messages don't necessarily have the same rates. The current implementation waits until all the messages arrive before returning OK, which causes the `NOT OK` status.

Yes, I mentioned this aspect here previously https://github.com/PX4/PX4-GPSDrivers/pull/58#pullrequestreview-465750152 and it is not so easy to solve this correctly in general case. 

> In the Ashtech implementation, every time that a valid position or speed is received, it returns 1. I testes this approach and now we get 10Hz rates for speed and velocity. Satellite information and heading update the _gps_position as they arrive.

You make a workaround. By ignoring correct grouping of nmea messages, the heading and error information in `vehicle_gps_position` messages may be delayed each time by one period. While this delay is no so big, but still at least for my opinion it is not right to do this like you did. 

Regarding this pull request to px4 as itself I would like to [repeat myself](https://github.com/PX4/PX4-GPSDrivers/pull/58#issuecomment-746144625):
For my POV, this driver should be GPSBaseStationSupport class for all nmea/nmea-like receivers including ashtech with possibility to send commands to different known receivers (in px4 currently only ashtech). But it needs someone's time who has px4 hardware and several receivers including ashtech. Sorry, for now can make only advises.

jajberni:
@lukegluke very good points and thank you for your feedback! 

> You make a workaround. By ignoring correct grouping of nmea messages, the heading and error information in vehicle_gps_position messages may be delayed each time by one period. While this delay is no so big, but still at least for my opinion it is not right to do this like you did.

This is right and I am now observing this in the bag files where the GPS topics come grouped in two (I'm returning 1 on valid position and velocity messages):
image

I guess it makes more sense to wait until you have position, velocity and error before returning a valid position. Satellite information is not that critical and I don't think it would cause any trouble to have a delay in the number of satellites. Heading is trickier because it doesn't even have a timestamp. 

I am not so sure about the GPSBaseStationSupport requirement because this is quite specific and I don't follow why this should be part of the PX4 driver. The base station support is meant to enable SurveryIn and transmission of RTCM messages for RTK correction. Why should this be implemented on the vehicle (rover) and not just on the base station? I haven't followed up on the discussions (I am sure there has to be a reason for it beyond the support of dual GNSS systems for heading). 

In my opinion, the astech driver is quite specific to the astech/trimble dialect, with heaps of non-standard messages. Maybe what is required is just a command for custom initialization (very common on generic GPS software) that is passed to the GPS, but only read/interpret standard NMEA sentences to keep it generic. Despite all the limitations of NMEA, most high-end GPS units can provide NMEA sentences. 

Also, how can you keep back compatibility with astech GPS type? The safer solution in the mid-term is to add a new type and eventually deprecate astech if both get merged.

jajberni:
> @jajberni can you open a pull request?

Let me do some more testing in the coming days and try to implement the comments from @lukegluke.

lukegluke:
> Satellite information is not that critical and I don't think it would cause any trouble to have a delay in the number of satellites. 

I agree about number of sats. Probably about errors too. But I really suppose that their should not be delaying in heading information, because it is separately get fused in ekf2 filter. 

> I am not so sure about the GPSBaseStationSupport requirement because this is quite specific and I don't follow why this should be part of the PX4 driver.

Here we are in PX4-GPSDrivers repository that is used not only for PX4-Autopilot, but also in QGC. And it exactly uses this drivers for working with RTK receivers as SurveryIn/Fixed base station. Maybe for now px4 community used limited number of well known receivers, I would like to recommend to already make enough scalable solution for future.

But also speaking about implementing RTCM messages on the vehicle (rover). It obviously useful to make post processing (PPK). I even wonder why it is not implemented in px4, while architecturally there is already a GPSCallbackType::gotRTCMMessage? I thought that PPK is essential to aerophotography and aerophotogrammetry.
And there is no so much work to implement this: I see this like handling gotRTCMMessage callback, publishing orb topic `gps_rtcm_data` (same as `gps_inject_data` - or they even be combined in same way as `gps_dump` using MSB bit), recording topic in logger to sd card.
 
> In my opinion, the astech driver is quite specific to the astech/trimble dialect, with heaps of non-standard messages. 

Agree, but still there is quite of common code that can be combined. Code in this PR was quite copied from astech.

> Also, how can you keep back compatibility with astech GPS type?

I suppose there can be implemented an autodetect by requesting particular astech get version command or something like this. But in first place if GPS_DRIVER_MODE_ASHTECH we can just pass additional specific nmea_flags in one general nmea driver (or specific type var) that is tell us that receiver is astech (or any other specific further).

> The safer solution in the mid-term is to add a new type and eventually deprecate astech if both get merged.

Or like this, yes.

jajberni:
I have created a pull request in https://github.com/PX4/PX4-GPSDrivers/pull/74 

Now the code works nicely at 10Hz and a valid `vehicle_gps_position` is generated only when we got a valid VEL and POS. The code checks the UTC time from the NMEA sentence and the last received timestamp to avoid duplicated information until the next NMEA block. 

Unfortunately, HDT and VTG don't have UTC information. I guess we'll need to live with that and update as they arrive. I am not sure if there are any NMEA order defined (I haven't found any clear indication) to determine if HDT is sent at the beginning or end of the NMEA block.

lukegluke:
> Unfortunately, HDT and VTG don't have UTC information.

Indeed, but they have gps_absolute_time on receive to distinguish the order (if assuming that output rate of these messages is the same as main ones GGA/RMC)

bkueng:
Added in https://github.com/PX4/PX4-GPSDrivers/pull/87
Please submit a PR if you need any change.

PR #60, opened 2020-10-13
Add heading for u-blox F9P

bkueng:


src/ubx.cpp
```diff
@@ -68,12 +68,15 @@
...
+			   sensor_gps_s *gps_position, satellite_info_s *satellite_info, uint8_t dynamic_model,
+			   float heading_offset, UBXMode mode) :
 	GPSBaseStationSupport(callback, callback_user),
 	_interface(gpsInterface),
 	_gps_position(gps_position),
 	_satellite_info(satellite_info),
-	_dyn_model(dynamic_model)
+	_dyn_model(dynamic_model),
+	_mode(mode),
+	_heading_offset(heading_offset)
```
mcsauder, line 79: Hi @bkueng , would you want to match the order of the constructor args? (Header file initialization order follows as well)

mcsauder:
Quick bench test with a SparkFun GPS-RTK2 ZED F9P, everything looks good to me.  Nice work.
image

src/ubx.cpp
bkueng: I don't think it needs to match, as you might have different reasons for odering each (e.g. constructor argument with default value). Only the initialization list needs to match the declaration order.

bkueng:
Thanks for testing

PR #61, opened 2020-11-02
replace non-ascii characters in source code

kemen209:
This will generate warnings to block the building process of QGC. See [related bug in QGC](https://github.com/mavlink/qgroundcontrol/pull/9143)

PR #62, opened 2020-11-28
The file use "UTF8 with BOM" encoding instead of "UTF8" encoding, com…

qsaker:
The file use "UTF8 with BOM" encoding instead of "UTF8" encoding, compile the project on Chinese system(Windows) will occur error when using UTF8 encoding.

### DIFF
--- a/src/ashtech.cpp
+++ b/src/ashtech.cpp
@@ -1,4 +1,4 @@
-/****************************************************************************
+﻿/****************************************************************************
  *
  *   Copyright (c) 2012-2018 PX4 Development Team. All rights reserved.
  *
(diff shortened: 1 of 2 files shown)


qsaker:
Oh, I use the QGC 4.0.11, the pull request is outmoded?

bkueng:
There was a PR (https://github.com/PX4/PX4-GPSDrivers/pull/61) that did this already, and your PR seems to undo that, so I'll close it.

I updated the QGC submodule in https://github.com/mavlink/qgroundcontrol/pull/9222. Can you let me know if that does not work for you?

PR #64, opened 2020-12-11
ubx: fix #63: NAV-SAT/ SVINFO parser bug

### LINKED ISSUE #63: uBlox Protocol Parser Issue with NAV-SAT and NAV-SVINFO
We are experiencing a mismatch between the data in the `satellite_info` uORB message (and the GPS_STATUS MAVLink message) and the UART data stream from our Pixhawk 4 GPS module. We analyzed the raw UART data stream by enabling the GPS_DUMP parameter in PX4 and adding the corresponding topic to the list of logged topics.

While comparing the code in the driver with the uBlox protocol specification from the M8N module (Release R20, 26 June 2020) I noticed some differences in the data format:
- The repeated structure of NAV-SAT and NAV-SVINFO should be 12 bytes long according to uBlox
- [ubx_payload_rx_nav_svinfo_part2_t](https://github.com/PX4/PX4-GPSDrivers/blob/master/src/ubx.h#L455) actually contains 14 bytes. Compared to uBlox:
  * `cno` should be before `elev`
  * there should be no SNR field
  * there should be no PRN field
- [ubx_payload_rx_nav_sat_part2_t](https://github.com/PX4/PX4-GPSDrivers/blob/master/src/ubx.h#L477) actually contains 13 bytes. Compared to uBlox:
  * `cno` should be at third place
  * there should be no PRN field
- Shouldn't these be packed structs? As far as I understand, you're filling the data by some pointer arithmetic, right?

This also leads to some issues with the `satellite_info` uORB message and [GPS_STATUS](https://mavlink.io/en/messages/common.html#GPS_STATUS) mavlink message. Currently the `satellite_prn` data of the MAVLink message is filled from the `prn` array of `satellite_info`.
But when this information is missing from the uBlox message:
- either the `satellite_info.prn` array should be filled with the ublox `svId` to ensure backwards compatibility
- or the MAVLink message should be using `satellite_info.svid` when `satellite_info.prn` is empty.

jbeyerstedt:
To maintain backwards compatibility with the GPS_STATUS MAVLink message, the `svId` was also added to `satellite_info.prn`.

Because `satellite_info.svid` is only 8 bit wide, the new `gnssId:svId` format used in the NAV-SAT message does not fit into the uORB topic. Therefore the data from NAV-SAT is converted to the old format from NAV-SVINFO.

### DIFF
--- a/src/ubx.cpp
+++ b/src/ubx.cpp
@@ -45,9 +45,9 @@
  * @author Hannes Delago
  *   (rework, add ubx7+ compatibility)
  *
- * @see https://www2.u-blox.com/images/downloads/Product_Docs/u-blox6-GPS-GLONASS-QZSS-V14_ReceiverDescriptionProtocolSpec_Public_(GPS.G6-SW-12013).pdf
+ * @see https://www.u-blox.com/sites/default/files/products/documents/u-blox6-GPS-GLONASS-QZSS-V14_ReceiverDescrProtSpec_%28GPS.G6-SW-12013%29_Public.pdf
  * @see https://www.u-blox.com/sites/default/files/products/documents/u-blox8-M8_ReceiverDescrProtSpec_%28UBX-13003221%29_Public.pdf
- * @see https://www.u-blox.com/sites/default/files/u-blox_ZED-F9P_InterfaceDescription_%28UBX-18010854%29.pdf
+ * @see https://www.u-blox.com/sites/default/files/ZED-F9P_InterfaceDescription_%28UBX-18010854%29.pdf
  */
 
 #include <string.h>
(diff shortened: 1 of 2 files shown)


src/ubx.cpp
```diff
@@ -1282,13 +1282,75 @@ GPSDriverUBX::payloadRxAddNavSat(const uint8_t b)
 				// Part 2 complete: decode Part 2 buffer
 				unsigned sat_index = (_rx_payload_index - sizeof(ubx_payload_rx_nav_sat_part1_t)) /
 						     sizeof(ubx_payload_rx_nav_sat_part2_t);
-				_satellite_info->svid[sat_index]	  = static_cast<uint8_t>(_buf.payload_rx_nav_sat_part2.svId);
+
+				// convert gnssId:svId to a 8 bit number (UBX svId from NAV-SVINFO, NMEA would need at last 9 bytes)
```
bkueng, line 1286: I guess you meant:
```suggestion
				// convert gnssId:svId to a 8 bit number (UBX svId from NAV-SVINFO, NMEA would need at least 9 bits)
```
And can you add where the numbering is coming from?

bkueng (commented):
Thanks.

The PRN field was recently added (https://github.com/PX4/PX4-GPSDrivers/pull/59). @mcsauder was this correct for you?

src/ubx.cpp
jbeyerstedt: Yes, it's a typo. I can also remove that bit of information.
I had the issue, that the NAV-SAT uses two 8 bit fields to define the satellite number, the GNSS ID (as "namespace") and the satellite's number (inside the namespace).
But satellite_info only has one 8 bit field. So there are 2 possible solutions: using the numbering scheme from the old NAV-SVINFO message or the NMEA numbering. But the NMEA numbering has numbers greater than 255, so it wouldn't fit. Therefore I used the "old" numbering.

mcsauder:
> Thanks.
> 
> The PRN field was recently added (#59). @mcsauder was this correct for you?
Hi @bkueng and @jbeyerstedt .  I validated that data was making it through the field, valid L1 CA/PRN values should be [120-144](https://www.gps.gov/technical/prn-codes/L1-CA-PRN-code-assignments-2020-Oct.pdf), but I didn't spend a lot of additional time evaluating why some values were not making complete sense, only that the data structure was getting populated and matching the existing structures and messages, #15871:
image

This PR looks like a good solution to me.  Thanks for digging into this @jbeyerstedt .

src/ubx.cpp
bkueng: Yes I understood that, and is fine for now, just wanted to make sure the comment makes that clear as well.

bkueng (approved):

Issue #68, opened 2021-01-06
Feature Request: Implement configuration of used GNSS systems


jbeyerstedt:
Many GPS receivers are also capable of receiving other GNSS systems, like GLONASS, Galileo and BeiDou. But it depends on the factory settings, which systems are enabled by default. It would be quite useful to have a way to configure the set of GNSS systems used.
See https://github.com/PX4/PX4-Autopilot/issues/16370 for details and discussion on how the parameter should be designed. 

This issue should keep track on the state of implementing this new PX4 parameter in all the GPS receiver drivers:

- [ ] Ashtech
- [ ] Emlid Reach
- [ ] Femtomes
- [ ] MTK
- [ ] SBF
- [ ] uBlox

PR #69, opened 2021-01-07
Implement new parameter for GNSS system selection

jbeyerstedt:
A new PX4 parameter is needed to configure the set of used GNSS systems. This PR changes the `GPSHelper::configure()` interface method used by the GPS drivers to include another parameter for the configuration (with the same data format, as the PX4 parameter).
Additionally the actual feature of setting the GNSS configuration was implemented for uBlox devices.

See https://github.com/PX4/PX4-GPSDrivers/issues/68 for details.

### DIFF
--- a/src/ashtech.cpp
+++ b/src/ashtech.cpp
@@ -942,9 +942,9 @@ int GPSDriverAshtech::waitForReply(NMEACommand command, const unsigned timeout)
 	return _command_state == NMEACommandState::received ? 0 : -1;
 }
 
-int GPSDriverAshtech::configure(unsigned &baudrate, OutputMode output_mode)
+int GPSDriverAshtech::configure(unsigned &baudrate, const GPSConfig &config)
 {
-	_output_mode = output_mode;
+	_output_mode = config.output_mode;
 	_correction_output_activated = false;
 	_configure_done = false;
 
@@ -1069,7 +1069,7 @@ int GPSDriverAshtech::configure(unsigned &baudrate, OutputMode output_mode)
 	// Enable dual antenna mode (2: both antennas are L1/L2 GNSS capable, flex mode, avoids the need to determine
 	// the baseline length through a prior calibration stage)
 	// Needs to be set before other commands
-	const bool use_dual_mode = output_mode == OutputMode::GPS && _board == AshtechBoard::trimble_mb_two;
+	const bool use_dual_mode = _output_mode == OutputMode::GPS && _board == AshtechBoard::trimble_mb_two;
 
 	if (use_dual_mode) {
 		ASH_DEBUG("Enabling DUO mode");
@@ -1118,7 +1118,7 @@ int GPSDriverAshtech::configure(unsigned &baudrate, OutputMode output_mode)
 	}
 
 
-	if (output_mode == OutputMode::RTCM && _board == AshtechBoard::trimble_mb_two) {
+	if (_output_mode == OutputMode::RTCM && _board == AshtechBoard::trimble_mb_two) {
 		SurveyInStatus status{};
 		status.latitude = status.longitude = (double)NAN;
 		status.altitude = NAN;
--- a/src/gps_helper.h
+++ b/src/gps_helper.h
@@ -161,6 +161,24 @@ public:
 		SPI
 	};
 
+	/**
+	 * Bitmask for GPS_1_GNSS and GPS_2_GNSS
+	 * No bits set should keep the receiver's default config
+	 */
+	enum class GNSSSystemsMask : int32_t {
+		RECEIVER_DEFAULTS = 0,
+		ENABLE_GPS =        1 << 0,
+		ENABLE_SBAS =       1 << 1,
+		ENABLE_GALILEO =    1 << 2,
+		ENABLE_BEIDOU =     1 << 3,
+		ENABLE_GLONASS =    1 << 4
+	};
+
+	struct GPSConfig {
+		OutputMode output_mode;
+		GNSSSystemsMask gnss_systems;
+	};
+
 
 	GPSHelper(GPSCallbackPtr callback, void *callback_user);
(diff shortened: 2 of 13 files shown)


jbeyerstedt:
I decided to not add a default value to the new parameter of that method, because it isn't needed any more when the corresponding change in PX4 was merged. But therefore the changes in this repository and in PX4 depend on each other. Let me know, if this is an issue.
My proposal for the procedure would be to merge this first and do the submodule update in PX4 as part of the https://github.com/PX4/PX4-Autopilot/pull/16505 pull request.

jbeyerstedt:
And could somebody please test this with a uBlox generation 9 receiver (or some other one, which is using at least protocol version 27/ the config-valset interface)? I only have a Pixhawk 4 GPS, which is using the `configureDevicePreV27()` code.

src/ashtech.cpp
```diff
@@ -942,7 +942,7 @@ int GPSDriverAshtech::waitForReply(NMEACommand command, const unsigned timeout)
 	return _command_state == NMEACommandState::received ? 0 : -1;
 }
 
-int GPSDriverAshtech::configure(unsigned &baudrate, OutputMode output_mode)
+int GPSDriverAshtech::configure(unsigned &baudrate, OutputMode output_mode, GNSSSystemsMask gnssSystems)
```
bkueng, line 945: As this is an interface change (and therefore requires all users of the drivers to update, so PX4 + QGC, and unknown ones), I'd rather create a separate struct to simplify future additions:
```suggestion
int GPSDriverAshtech::configure(unsigned &baudrate, const GPSConfig& config)
```
```
struct GPSConfig {
  OutputMode output_mode = xy;
  GNSSSystemsMask gnss_systems = xy;
};

File: src/gps_helper.h [hunk 2]
```diff
@@ -161,6 +161,18 @@ class GPSHelper
 		SPI
 	};
 
+	/**
+	 * Bitmask for GPS_1_GNSS and GPS_2_GNSS
+	 * No bits set should keep the receiver's default config
+	 */
+	enum class GNSSSystemsMask : int32_t {
```
bkueng, line 168: I'd add the `RECEIVER_DEFAULTS = 0` as well

src/ubx.cpp
```diff
@@ -89,10 +89,11 @@ GPSDriverUBX::~GPSDriverUBX()
 }
 
 int
-GPSDriverUBX::configure(unsigned &baudrate, OutputMode output_mode)
+GPSDriverUBX::configure(unsigned &baudrate, OutputMode output_mode, GNSSSystemsMask gnssSystems)
 {
 	_configured = false;
 	_output_mode = output_mode;
+	_gnssSystems = gnssSystems;
```
bkueng, line 96: I'd rather pass this to the method, as there's no need to keep it stored.

src/ubx.cpp
```diff
@@ -472,6 +547,64 @@ int GPSDriverUBX::configureDevice()
...
+
+		if (_gnssSystems & GNSSSystemsMask::ENABLE_BEIDOU) {
+			UBX_DEBUG("GNSS Systems: Use BeiDou");
+			cfgValset<uint8_t>(UBX_CFG_KEY_SIGNAL_BDS_ENA, 1, cfg_valset_msg_size);
+
+		} else {
+			cfgValset<uint8_t>(UBX_CFG_KEY_SIGNAL_BDS_ENA, 0, cfg_valset_msg_size);
+		}
+
+		_buf.payload_tx_cfg_gnss.block[5].gnssId = UBX_TX_CFG_GNSS_GNSSID_GLONASS;
```
bkueng, line 592:
```suggestion
```

bkueng (commented):
Thanks, I'll test this on a newer module.

src/gps_helper.h
jbeyerstedt: I already thought of this, but somehow didn't commit it. Good point.

src/ashtech.cpp
jbeyerstedt: Initially I thought of using a default value for the parameter, so it doesn't break things. But then I left it out because I haven't had in mind, that other projects than PX4 will use this repo.
But you suggestion is actually the cleaner and more future proof approach, since we breaks things with this anyways.

src/ubx.cpp
```diff
@@ -472,6 +546,62 @@ int GPSDriverUBX::configureDevice()
...
+
+		} else {
+			cfgValset<uint8_t>(UBX_CFG_KEY_SIGNAL_GLO_ENA, 0, cfg_valset_msg_size);
+		}
+
+		if (!sendMessage(UBX_MSG_CFG_VALSET, (uint8_t *)&_buf, cfg_valset_msg_size)) {
+			return -1;
+		}
+
+		waitForAck(UBX_MSG_CFG_VALSET, UBX_CONFIG_TIMEOUT, false);
```
bkueng, line 602:
```suggestion
		waitForAck(UBX_MSG_CFG_VALSET, UBX_CONFIG_TIMEOUT, true);
```

bkueng:
I tested with an F9P, but it does not work, I always get a NACK: setting `UBX_CFG_KEY_SIGNAL_SBAS_ENA` consistently fails, either enabling or disabling. The same result with u-center, so I think the code is correct, and I don't know why it fails. We could apply SBAS separately so at least the other settings apply.

I also noticed a bug: https://github.com/PX4/PX4-GPSDrivers/blob/master/src/ubx.cpp#L1294 should use `ubx_sat_svId`.

jbeyerstedt:
The SBAS settings have been moved to the bottom of the section, committing the other changes beforehand. Additionally I explicitly set the used signal for SBAS now (even tough it should be set to that value by default). Maybe this works better.

The other bug is addressed im https://github.com/PX4/PX4-GPSDrivers/pull/70.

bkueng (approved):

PR #71, opened 2021-01-29
Enable jamming and interference monitor 

cmic0:
title self explanatory, this should provide more insightful information about GPS jamming state.

From u-blox Integration Manual

> The field flags of the UBX-MON-RF message can be used as an indicator for both broadband andcontinuous wave (CW) jammers/interference. It is independent of the (CW only) jamming indicatordescribed in Jamming/interference indicator above.This monitor reports whether jamming has been detected or suspected by the receiver. The receivermonitors the background noise and looks for significant changes. Normally, with no interferencedetected, it will report "OK". If the receiver detects that the noise has risen above a preset threshold,the  receiver  reports  "Warning".  If  in  addition,  there  is  no  current  valid  fix,  the  receiver  reports"Critical".

### DIFF
--- a/src/ubx.h
+++ b/src/ubx.h
@@ -309,6 +309,8 @@
 #define UBX_CFG_KEY_ODO_OUTLPVEL                0x10220003
 #define UBX_CFG_KEY_ODO_OUTLPCOG                0x10220004
 
+#define UBX_CFG_KEY_ITFM_ENABLE					0x1041000d
+
 #define UBX_CFG_KEY_RATE_MEAS                   0x30210001
 #define UBX_CFG_KEY_RATE_NAV                    0x30210002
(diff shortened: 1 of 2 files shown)


src/ubx.h
```diff
@@ -309,6 +309,8 @@
 #define UBX_CFG_KEY_ODO_OUTLPVEL                0x10220003
 #define UBX_CFG_KEY_ODO_OUTLPCOG                0x10220004
 
+#define UBX_CFG_KEY_ITFM_ENABLE					0x1041000d
```
bkueng, line 312: Indentation

cmic0: done

bkueng (approved):
Note that this will need a QGC update as well when updating the submodule, but I can do that.

PR #74, opened 2021-03-03
Add Generic NMEA Protocol GPS Driver

jajberni:
Based on DimianZhan:add_nmea_driver and the PR https://github.com/PX4/PX4-GPSDrivers/pull/58

This code complies with the new `configure()` function and changes the default behaviour for accepting valid `vehicle_gps_position`.

The code waits until valid velocity and position are achieved (`_VEL_received && _POS_received`) in some of the NMEA sentences. To avoid duplicated information, the utc_time of the last POS and VEL are checked:

```
		if (!_POS_received && (_last_POS_timeUTC < utc_time)) {
			_last_POS_timeUTC = utc_time;
			_POS_received = true;
		}
		if (!_VEL_received && (_last_VEL_timeUTC < utc_time)) {
			_last_VEL_timeUTC = utc_time;
			_VEL_received = true;
		}
```
Unfortunately, messages like HDT or VTG don't contain the UTC time, so it is not possible to link them to the NMEA sentence block at each time. 

Heading and satellite information are updated as they arrive, but don't trigger a new valid `vehicle_gps_position`.

The code has been tested with a NorthSurverying RTKite with a dual antenna and we get 10Hz updates with heading info. 

As discussed in https://github.com/PX4/PX4-GPSDrivers/pull/58 I am not sure if the generic NMEA should derive from GPSBaseStationSupport, as this is highly specific to each vendor, while the spirit of NMEA is being as generic as possible.

### DIFF
--- a/src/nmea.cpp
+++ b/src/nmea.cpp
@@ -767,7 +767,6 @@ int GPSDriverNMEA::handleMessage(int len)
 			}
 		}
 
-		// mavlink_log_info(&mavlink_log_pub, "get GSV data ");
 
 	} else if ((memcmp(_rx_buffer + 3, "VTG,", 4) == 0) && (uiCalcComma >= 8)) {
 
--- a/src/nmea.cpp
+++ b/src/nmea.cpp
@@ -290,6 +290,8 @@ int GPSDriverNMEA::handleMessage(int len)
 
 		_gps_position->c_variance_rad = 0.1f;
 		_gps_position->timestamp = gps_absolute_time();
+		_rate_count_lat_lon++;
+		ret = 1;
 
 		// mavlink_log_info(&mavlink_log_pub, "GGA time->>> %d ",(int)(utc_time));
 
@@ -490,9 +492,9 @@ int GPSDriverNMEA::handleMessage(int len)
 		int utc_hour = static_cast<int>(utc_time / 10000);
 		int utc_minute = static_cast<int>((utc_time - utc_hour * 10000) / 100);
 		double utc_sec = static_cast<double>(utc_time - utc_hour * 10000 - utc_minute * 100);
-		int nema_day = static_cast<int>(nmea_date / 10000);
-		int nmea_mth = static_cast<int>((nmea_date - nema_day * 10000) / 100);
-		int nmea_year = static_cast<int>(nmea_date - nema_day * 10000 - nmea_mth * 100);
+		int nmea_day = static_cast<int>(nmea_date / 10000);
+		int nmea_mth = static_cast<int>((nmea_date - nmea_day * 10000) / 100);
+		int nmea_year = static_cast<int>(nmea_date - nmea_day * 10000 - nmea_mth * 100);
 		/* convert from degrees, minutes and seconds to degrees */
 		_gps_position->lat = static_cast<int>((int(lat * 0.01) + (lat * 0.01 - int(lat * 0.01)) * 100.0 / 60.0) * 10000000);
 		_gps_position->lon = static_cast<int>((int(lon * 0.01) + (lon * 0.01 - int(lon * 0.01)) * 100.0 / 60.0) * 10000000);
@@ -504,7 +506,9 @@ int GPSDriverNMEA::handleMessage(int len)
 		_gps_position->vel_ned_valid = true; /**< Flag to indicate if NED speed is valid */
 		_gps_position->c_variance_rad = 0.1f;
 		_gps_position->s_variance_m_s = 0;
+		_gps_position->timestamp = gps_absolute_time();
 		_last_VEL_timeUTC = utc_time;
+		_last_timestamp_time = gps_absolute_time();
 
 		/*
 		 * convert to unix timestamp
@@ -512,7 +516,7 @@ int GPSDriverNMEA::handleMessage(int len)
 		struct tm timeinfo = {};
 		timeinfo.tm_year = nmea_year + 100;
 		timeinfo.tm_mon = nmea_mth - 1;
-		timeinfo.tm_mday = nema_day;
+		timeinfo.tm_mday = nmea_day;
 		timeinfo.tm_hour = utc_hour;
 		timeinfo.tm_min = utc_minute;
 		timeinfo.tm_sec = int(utc_sec);
@@ -550,6 +554,8 @@ int GPSDriverNMEA::handleMessage(int len)
 		_POS_received = true;
 		_TIME_received = true;
 		_VEL_received = true;
+		_rate_count_vel++;
+		ret = 1;
 
 		// mavlink_log_info(&mavlink_log_pub, "RMC time->>> %d ",(int)(utc_time));
 
@@ -847,48 +853,26 @@ int GPSDriverNMEA::handleMessage(int len)
 		_gps_position->s_variance_m_s = 0;
 
 		_VEL_received = true;
+		_rate_count_vel++;
+		ret = 1;
 		// mavlink_log_info(&mavlink_log_pub, "get VTG data ");
 	}
 
-	if (_POS_received &&
-	    _ALT_received &&
-	    _SVNUM_received &&
+	if (_SVNUM_received &&
 	    _SVINFO_received &&
-	    _FIX_received &&
-	    (_DOP_received ||
-	     _EPH_received) &&
-	    _VEL_received &&
-	    _TIME_received &&
-	    ((_decode_flags & (int)NMEADecodeFlags::got_hdt) ? _HEAD_received : true)) {
+	    _FIX_received
+	    ) {
 
 		sat_num_gsv = sat_num_gpgsv + sat_num_glgsv + sat_num_gagsv
 			      + sat_num_gbgsv + sat_num_bdgsv;
 
-		_rate_count_lat_lon++;
-		_rate_count_vel++;
-
 		_gps_position->satellites_used = MAX(sat_num_gns, sat_num_gsv);
 		_gps_position->satellites_used = MAX(_gps_position->satellites_used, sat_num_gga);
+	}
 
+	if (ret == 1) {
+		_gps_position->timestamp_time_relative = (int32_t)(_last_timestamp_time - _gps_position->timestamp);
 		clock_set = false;
-
-		_TIME_received = false;
-		_POS_received = false;
-		_ALT_received = false;
-		_SVNUM_received = false;
-		_SVINFO_received = false;
-		_FIX_received = false;
-		_DOP_received = false;
-		_VEL_received = false;
-		_EPH_received = false;
-		_HEAD_received = false;
-		sat_num_gpgsv = 0;
-		sat_num_glgsv = 0;
-		sat_num_gagsv = 0;
-		sat_num_gbgsv = 0;
-		sat_num_bdgsv = 0;
-
-		ret = 1;
 	}
 
 	return ret;
--- a/src/nmea.cpp
+++ b/src/nmea.cpp
@@ -38,6 +38,7 @@
  *
  * @author WeiPeng Guo <guoweipeng1990@sina.com>
  * @author Stone White <stone@thone.io>
+ * @author Jose Jimenez-Berni <berni@ias.csic.es>
  *
  */
 
@@ -265,7 +266,7 @@ int GPSDriverNMEA::handleMessage(int len)
 		_gps_position->alt = static_cast<int>(alt * 1000);
 		_gps_position->alt_ellipsoid = _gps_position->alt + static_cast<int>(geoid_h * 1000);
 		sat_num_gga = static_cast<int>(num_of_sv);
-		_last_POS_timeUTC = utc_time;
+
 
 		if (fix_quality <= 0) {
 			_gps_position->fix_type = 0;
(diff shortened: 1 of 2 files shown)


src/nmea.cpp
```diff
@@ -0,0 +1,1043 @@
...
+		if (ns == 'S') {
+			lat = -lat;
+		}
+
+		if (ew == 'W') {
+			lon = -lon;
+		}
+
+		if (Status == 'V') {
+			return 0;
```
lukegluke, line 484: I don't think it is right to return from RMC there is another information in it below, I suggest to just clear fix_type without return:
`_gps_position->fix_type = 0;`

src/nmea.h
```diff
@@ -0,0 +1,138 @@
...
+	bool _DOP_received{false};
+	bool _VEL_received{false};
+	bool _EPH_received{false};
+	bool _HEAD_received{false};
+
+	NMEADecodeState _decode_state{NMEADecodeState::uninit};
+	uint8_t _rx_buffer[NMEA_RECV_BUFFER_SIZE] {};
+	uint16_t _rx_buffer_bytes{0};
+
+	uint8_t _read_buf[GPS_READ_BUFFER_SIZE] {};
```
bkueng, line 133: Can you keep this as local variable in the method and parse the whole buffer, so that stack & memory usage are roughly the same for all drivers?

src/nmea.cpp
```diff
@@ -0,0 +1,1043 @@
...
+	GPSHelper(callback, callback_user),
+	_gps_position(gps_position),
+	_satellite_info(satellite_info),
+	_heading_offset(heading_offset)
+{
+	decodeInit();
+}
+
+/*
+ * All NMEA descriptions are taken from
```
bkueng, line 73: Is this accurate?

src/nmea.cpp
```diff
@@ -0,0 +1,1043 @@
...
+	_decode_state = NMEADecodeState::uninit;
+}
+
+int GPSDriverNMEA::configure(unsigned &baudrate, const GPSConfig &config)
+{
+	if (config.output_mode != OutputMode::GPS) {
+		GPS_WARN("NMEA: Unsupported Output Mode %i", (int)config.output_mode);
+		return -1;
+	}
+	setBaudrate(baudrate);
```
bkueng, line 1016: `baudrate` might be 0, in which case there should be auto-detection. If not supported, you can just set `baudrate` to `NMEA_DEFAULT_BAUDRATE` if it's 0.

src/nmea.h
```diff
@@ -0,0 +1,138 @@
...
+	uint8_t sat_num_gns{0};
+	uint8_t sat_num_gsv{0};
+	uint8_t sat_num_gpgsv{0};
+	uint8_t sat_num_glgsv{0};
+	uint8_t sat_num_gagsv{0};
+	uint8_t sat_num_gbgsv{0};
+	uint8_t sat_num_bdgsv{0};
+
+	//nmea protocol decode flags NMEADecodeFlags
+	uint32_t _decode_flags{0};
```
bkueng, line 113: Do you intend to use this for something?

src/nmea.h
```diff
@@ -0,0 +1,138 @@
...
+	char read_char();
+
+	sensor_gps_s *_gps_position {nullptr};
+	satellite_info_s *_satellite_info {nullptr};
+	uint32_t _last_POS_timeUTC{0};
+	uint32_t _last_VEL_timeUTC{0};
+	uint32_t _last_FIX_timeUTC{0};
+	uint64_t _last_timestamp_time{0};
+
+	uint8_t sat_num_gga{0};
```
bkueng, line 103:
```suggestion
	uint8_t _sat_num_gga{0};
```
Other as well

src/nmea.h
```diff
@@ -0,0 +1,138 @@
...
+	uint8_t sat_num_gpgsv{0};
+	uint8_t sat_num_glgsv{0};
+	uint8_t sat_num_gagsv{0};
+	uint8_t sat_num_gbgsv{0};
+	uint8_t sat_num_bdgsv{0};
+
+	//nmea protocol decode flags NMEADecodeFlags
+	uint32_t _decode_flags{0};
+
+	bool clock_set {false};
```
bkueng, line 115:
```suggestion
	bool _clock_set {false};
```

src/nmea.cpp
```diff
@@ -0,0 +1,1043 @@
...
+		_gps_position->vel_n_m_s = velocity_north;
+		_gps_position->vel_e_m_s = velocity_east;
+		_gps_position->cog_rad = track_rad;
+		_gps_position->vel_ned_valid = true; /** Flag to indicate if NED speed is valid */
+		_gps_position->c_variance_rad = 0.1f;
+		_gps_position->s_variance_m_s = 0;
+
+		if (!_VEL_received) {
+			_VEL_received = true;
+			_rate_count_vel++;
```
bkueng, line 864: This can be double counted
```suggestion
```

bkueng (commented):
> As discussed in #58 I am not sure if the generic NMEA should derive from GPSBaseStationSupport, as this is highly specific to each vendor, while the spirit of NMEA is being as generic as possible.

I'd rather have a separate NMEA parser class, to parse messages like `GGA`. I haven't checked in detail, but there's not so much overlap to the ashtech driver.

src/nmea.cpp
jajberni: Good idea and follows the same approach as in other quality indicators.

src/nmea.cpp
jajberni: This comes from the original source code developed by DimianZhan:add_nmea_driver I am not sure how to do the right author attribution (as I couldn't branch from that one).

src/nmea.h
jajberni: Not, really, this was also inherited from the original code and now we check the status in a different way. It is ok to remove it.

src/nmea.cpp
jajberni: I had some problems with the original code and removed the auto-detection. I will try again and test the code or add your suggestions.

jajberni: Now the automatic detection of baudrate works if `baudrate` is 0. Also, the fallback is the `NMEA_DEFAULT_BAUDRATE` if no GPS is detected (except if specified).

src/nmea.h
jajberni: Moved to the local function as in the astech driver.

src/nmea.h
```diff
@@ -0,0 +1,131 @@
...
+		      sensor_gps_s *gps_position,
+		      satellite_info_s *satellite_info,
+		      float heading_offset = 0.f);
+	virtual ~GPSDriverNMEA() = default;
+
+	int receive(unsigned timeout) override;
+	int configure(unsigned &baudrate, const GPSConfig &config) override;
+
+private:
+	enum class NMEADecodeFlags : int {
```
bkueng, line 69: This can be removed as well

src/nmea.cpp
bkueng: I would just remove it then, as it's likely a copy-paste leftover and possibly add your reference

src/nmea.cpp
```diff
@@ -0,0 +1,1041 @@
...
+
+		if (bufptr && *(++bufptr) != ',') { lon = strtod(bufptr, &endp); bufptr = endp; }
+
+		if (bufptr && *(++bufptr) != ',') { ew = *(bufptr++);}
+
+		do {
+			pos_Mode[i] =  *(bufptr);
+			i++;
+
+		} while (*(++bufptr) != ',');
```
bkueng, line 377: just to be safe:
```suggestion
		} while (*(++bufptr) != ',' && i < 5);
```

src/nmea.cpp
```diff
@@ -0,0 +1,1041 @@
...
+			_last_POS_timeUTC = utc_time;
+			_POS_received = true;
+		}
+		_ALT_received = true;
+		_SVNUM_received = true;
+
+		// mavlink_log_info(&mavlink_log_pub, "GNS> posMode %s/sv_num %d/ hdop %.2f",
+		// (char*)(pos_Mode),
+		// (int)(sat_num_gns),
+		// (double)(hdop));
```
bkueng, line 410:
```suggestion
```

src/nmea.cpp
```diff
@@ -0,0 +1,1041 @@
...
+
+	}	else if ((memcmp(_rx_buffer + 3, "RMC,", 4) == 0) && (uiCalcComma >= 11)) {
```
bkueng, line 412: Did you run the style check from PX4 Autopilot?
```suggestion
	} else if ((memcmp(_rx_buffer + 3, "RMC,", 4) == 0) && (uiCalcComma >= 11)) {
```

src/nmea.cpp
```diff
@@ -0,0 +1,1041 @@
...
+			_gps_position->fix_type = 0;
+
+		} else {
+			_gps_position->hdop = static_cast<float>(hdop);
+			_gps_position->vdop = static_cast<float>(vdop);
+			_DOP_received = true;
+
+		}
+
+		// mavlink_log_info(&mavlink_log_pub, "gsa data->>>hdop %.2f---vdop %.2f ",(double)(_gps_position->hdop),(double)(_gps_position->vdop));
```
bkueng, line 675:
```suggestion
```

src/nmea.cpp
```diff
@@ -0,0 +1,1041 @@
...
+		*/
+
+		int all_page_num = 0, this_page_num = 0, tot_sv_visible = 0;
+		struct gsv_sat {
+			int svid;
+			int elevation;
+			int azimuth;
+			int snr;
+		} sat[4];
+		memset(sat, 0, sizeof(sat));
```
bkueng, line 706:
```suggestion
		} sat[4] {};
```

src/nmea.cpp
```diff
@@ -0,0 +1,1041 @@
...
+
+				_satellite_info->svid[y + (this_page_num - 1) * 4]      = sat[y].svid;
+				_satellite_info->used[y + (this_page_num - 1) * 4]      = (sat[y].snr > 0);
+				_satellite_info->snr[y + (this_page_num - 1) * 4]       = sat[y].snr;
+				_satellite_info->elevation[y + (this_page_num - 1) * 4] = sat[y].elevation;
+				_satellite_info->azimuth[y + (this_page_num - 1) * 4]   = sat[y].azimuth;
+			}
+		}
+
+		// mavlink_log_info(&mavlink_log_pub, "get GSV data ");
```
bkueng, line 775:
```suggestion
```

src/nmea.cpp
```diff
@@ -0,0 +1,1041 @@
...
+			int l = 0;
+
+			if ((l = parseChar(buf[j])) > 0) {
+				/* return to configure during configuration or to the gps driver during normal work
+				 * if a packet has arrived */
+				int ret = handleMessage(l);
+
+				if (ret > 0) {
+					return ret;
+				}
```
bkueng, line 898: Looking at this now, there could still be more data. The ubx driver handles this correctly.

bkueng (commented):
Looks pretty good

src/nmea.h
jajberni: Sorry, I missed this one!

src/nmea.cpp
jajberni: I'll look into the ubx driver then. I am not so confident fine-tuning these lower lever functions.

bkueng: @jajberni any update? I can also look into it if you want me to.

lukegluke: Just in case this issue was already mentioned in previous discussion https://github.com/PX4/PX4-GPSDrivers/pull/58, so if it be usefull:

```
/* pass received bytes to the packet decoder */
while (_bytes_parsed < _bytes_readed) {
	int l = 0;

	if ((l = parseChar(_read_buf[_bytes_parsed])) > 0) {
		/* return to configure during configuration or to the gps driver during normal work
			* if a packet has arrived */
		if (handleMessage(l) > 0) {
			return 1;
		}
	}

	_bytes_parsed++;
}

/* everything is read */
_bytes_parsed = _bytes_readed = 0;
```

jajberni: Sorry, I have been sidetracked with other aspects of the project and haven't looked into this yet. Let me have a look and test it. If I don't find the time or any suitable solution, I will be more than happy to accept your help! :-)

bkueng: @jajberni I take it you didn't have time so I did the updates in https://github.com/PX4/PX4-GPSDrivers/tree/generic_nmea_driver, can you test it?

nrodichenko:
@bkueng @jajberni Are there any modifications that need to be done prior to merging?

bkueng:
@nrodichenko No, I made a couple of additional changes in https://github.com/PX4/PX4-GPSDrivers/tree/generic_nmea_driver. If you can test that branch and confirm it's working, we can get that merged.

nrodichenko:
Ok, will try to test soon. Is there a px4 branch already setup for that?

bkueng:
> Ok, will try to test soon. Is there a px4 branch already setup for that?

Only a very old one with conflicts and some of the code got implemented already: https://github.com/PX4/PX4-Autopilot/pull/14773/files

jajberni:
Sorry for the delay, we have been sidetracked and haven't had time to test the code with the changes. Unfortunately, we are not getting heading with the current code (it reports nan). I guess that returning 1 with only position and velocity skips the heading information. It would be necessary to test if we are receiving heading messages, and in this case, wait to get the HDT before returning 1. Any suggestion? I'll keep trying in the meantime.

bkueng:
`_gps_position->heading` is kept until the next position is published. But if it comes in at a lower rate, which is typically the case, not all position messages will contain the heading, and you have to check all messages over a period of time.

bkueng:
Added in https://github.com/PX4/PX4-GPSDrivers/pull/87

PR #75, opened 2021-03-10
Adding messageId function to return what RTCM message is in the buffer.

TobinHall:
Fairly simple function to pull the RTCM message id out of the buffer. 
Useful for message filtering.

### DIFF
--- a/src/rtcm.h
+++ b/src/rtcm.h
@@ -60,6 +60,7 @@ public:
 
 	uint8_t *message() const { return _buffer; }
 	uint16_t messageLength() const { return _pos; }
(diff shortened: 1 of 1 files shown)


src/rtcm.h
```diff
@@ -60,6 +60,7 @@ class RTCMParsing
 
 	uint8_t *message() const { return _buffer; }
 	uint16_t messageLength() const { return _pos; }
+	uint16_t messageId() const { return (_buffer[3]<<4)|(_buffer[4]>>4); }
```
bkueng, line 63: Style check will probably complain in PX4:
```suggestion
	uint16_t messageId() const { return (_buffer[3] << 4) | (_buffer[4] >> 4); }
```

bkueng (approved):

PR #76, opened 2021-03-16
add GPS+RTCM output mode for logging (for PPK)

bkueng:


LorenzMeier (approved):

LorenzMeier:
Thanks!

src/ubx.cpp
```diff
@@ -83,9 +83,7 @@ GPSDriverUBX::GPSDriverUBX(Interface gpsInterface, GPSCallbackPtr callback, void
 
 GPSDriverUBX::~GPSDriverUBX()
 {
-	if (_rtcm_parsing) {
-		delete (_rtcm_parsing);
-	}
+	delete _rtcm_parsing;
```
lukegluke, line 86: @bkueng it looks like the bug. `_rtcm_parsing` is not created in simple GPS OutputMode. It is not a good idea to remove `_rtcm_parsing` null checking.

bkueng: Passing `nullptr` to `delete` is valid

lukegluke: Yeah, you are right, I missed it. Then it also could be done in other drivers, too :)

bkueng: Yes, definitely

PR #77, opened 2021-03-16
ubx log automatic gain control (agc) monitor

dagar:


LorenzMeier (approved):

LorenzMeier:
The change looks ok, but a quick search suggests that you haven't extended the struct with the actual field?

bkueng (approved):
Struct def is coming from the firmware

PR #80, opened 2021-04-27
update femtomes gps driver code for both px4 and QGC

Femtomes:
1. add base station driver code for QGC
     a. support RTK GPS with RTCM3
     b. support show satellite count used in base station
2. fix buid warning of px4


src/femtomes.cpp
```diff
@@ -103,17 +120,97 @@ int GPSDriverFemto::handleMessage(int len)
...
+                lat = (int(lat * 0.01) + (lat * 0.01 - int(lat * 0.01)) * 100.0 / 60.0) * 10000000;
+                lon = (int(lon * 0.01) + (lon * 0.01 - int(lon * 0.01)) * 100.0 / 60.0) * 10000000;
+                alt = alt * 1000;
+
+                activateRTCMOutput();
+                sendSurveyInStatusUpdate(false, true, lat, lon, (float)alt);
+                _correction_output_activated = true;
+            }
+
+            _satellite_info->count = (uint8_t)num_of_sv;    /**< base station satellite count */
```
bkueng, line 209:
```suggestion
           if (_satellite_info) {
            _satellite_info->count = (uint8_t)num_of_sv;    /**< base station satellite count */
            }
```

src/femtomes.cpp
```diff
@@ -103,17 +120,97 @@ int GPSDriverFemto::handleMessage(int len)
...
 
 		ret = 1;
 	}
+	else if(messageid == FEMTO_MSG_ID_RTCM3)    /**< rtcm3 */
+    {
+        gotRTCMMessage(_rtcm_parsing->message(), _rtcm_parsing->messageLength());
+        decodeInit();
+        ret = 1;
+    }
+	else if(messageid == FEMTO_MSG_ID_GPGGA && (memcmp(_femto_msg.data + 3, "GGA,", 3) == 0)) /**< GPGGA */
```
bkueng, line 150: Is this only ever received by the base station?
If it's possible to receive this on the vehicle side as well, `activateRTCMOutput` should not be called (you can add a check for `_output_mode`).

src/femtomes.cpp
```diff
@@ -189,116 +286,213 @@ int GPSDriverFemto::receive(unsigned timeout)
...
 	_decode_state = FemtoDecodeState::pream_ble1;
-
+    /** init or reset rtcm parsing */
+    if(_output_mode == OutputMode::RTCM)
+    {
+        if(!_rtcm_parsing)
+        {
+            _rtcm_parsing = new RTCMParsing();
+        }
+        _rtcm_parsing->reset();
```
bkueng, line 494:
```suggestion
        if(_rtcm_parsing) {
        _rtcm_parsing->reset();
        }
```

src/femtomes.cpp
```diff
@@ -434,3 +639,101 @@ GPSDriverFemto::calculateBlockCRC32(uint32_t length, uint8_t *buffer, uint32_t c
 	return (crc);
 }
 
+void GPSDriverFemto::activateCorrectionOutput()
+{
+    if (_correction_output_activated || _output_mode != OutputMode::RTCM)
```
bkueng, line 644: You shouldn't have to check for `_correction_output_activated`, as this is called once during configure.
```suggestion
    if (_output_mode != OutputMode::RTCM)
```

src/femtomes.cpp
```diff
@@ -434,3 +639,101 @@ GPSDriverFemto::calculateBlockCRC32(uint32_t length, uint8_t *buffer, uint32_t c
...
+            {
+                FEMTO_ERR("Femto: LOG GPGGA command failed")
+            }
+        }
+        else
+        {
+            FEMTO_ERR("Femto: command POSAVE ON failed")
+        }
+
+        _base_settings.settings.survey_in.min_dur = 0; // use it as counter how long survey-in has been active
```
bkueng, line 673: Where is this incremented? It should be updated once per second, which is displayed in QGC

src/femtomes.cpp
```diff
@@ -434,3 +639,101 @@ GPSDriverFemto::calculateBlockCRC32(uint32_t length, uint8_t *buffer, uint32_t c
...
+        return;
+    }
+
+    char buffer[100];
+
+    if (_base_settings.type == BaseSettingsType::survey_in)
+    {
+        FEMTO_DEBUG("Femto: enabling survey-in")
+
+        if(writeAckedCommandFemto("POSAVE ON \r\n","<POSAVE OK",FEMTO_RESPONSE_TIMEOUT) == 0)
```
bkueng, line 655: How long will the survey-in take and is that configurable?

src/femtomes.cpp
```diff
@@ -103,17 +120,97 @@ int GPSDriverFemto::handleMessage(int len)
...
+
+            if(!_correction_output_activated && 7 == fix_quality)
+            {
+                lat = (int(lat * 0.01) + (lat * 0.01 - int(lat * 0.01)) * 100.0 / 60.0) * 10000000;
+                lon = (int(lon * 0.01) + (lon * 0.01 - int(lon * 0.01)) * 100.0 / 60.0) * 10000000;
+                alt = alt * 1000;
+
+                activateRTCMOutput();
+                sendSurveyInStatusUpdate(false, true, lat, lon, (float)alt);
+                _correction_output_activated = true;
```
bkueng, line 206: Can you set this in `activateRTCMOutput`?

bkueng (commented):
Thanks. Can you run `make format` from the px4 firmware?

src/femtomes.cpp
Femtomes: Thank you ! You are right, I should add a check for _output_mode here. Although it's only for base stations

src/femtomes.cpp
Femtomes: Ok, I will set this in activateRTCMOutput.

src/femtomes.cpp
Femtomes: Hi bkueng ! I don't understand this suggestion.

src/femtomes.cpp
Femtomes: It will take a few seconds. The command POSAVE ON will reply immediately,And device will set the average position of the base station, and the position average will continue until fixed, then start RTCM3 output.

src/femtomes.cpp
Femtomes: Thanks for reminding! I forgot to implement this. I will implement it later.

Femtomes:
I have passed test in both px4_fmu_v5_default and QGC

src/femtomes.cpp
```diff
@@ -434,3 +657,102 @@ GPSDriverFemto::calculateBlockCRC32(uint32_t length, uint8_t *buffer, uint32_t c
...
+        int len = snprintf(buffer, sizeof(buffer), "FIX POSITION %.8lf %.8lf %.5f\r\n",
+                   settings.latitude, settings.longitude,(double)settings.altitude);
+        if (len >= 0 && len < (int)(sizeof(buffer)))
+        {
+            if (writeAckedCommandFemto(buffer, "FIX OK", FEMTO_RESPONSE_TIMEOUT) == 0)
+            {
+                FEMTO_DEBUG("Femto: command %s success", buffer)
+                activateRTCMOutput();
+                sendSurveyInStatusUpdate(false, true, settings.latitude, settings.longitude, settings.altitude);
+                _correction_output_activated = true;
```
bkueng, line 709:
```suggestion
```

src/femtomes.cpp
bkueng: So the averaging duration cannot be configured?

src/femtomes.cpp
bkueng: The allocation might fail if out of memory (not a problem on the GCS, but on the vehicle), so we need to check again if not null.

bkueng (commented):
Can you run `make format` from the px4 firmware? Let me know if you have problems with that.

src/femtomes.cpp
Femtomes: Yes，this is currently not configurable

Femtomes:
> Can you run `make format` from the px4 firmware? Let me know if you have problems with that.

Sorry, I didn't understand what you meant before, I have run make format，and pass QGC and px4 test

bkueng (approved):
Thanks for the quick updates. Looks good now.

PR #81, opened 2021-05-06
Disable GPS protocols at I2C if communication is going over UART1

Igor-Misic:
### Problem this PR solves

It is observed when the Ublox ZED-F9P module is connected to FMU over UART1 wires + I2C wires (that is the case if it is connected over GPS connector) it will stop sending **UBX_MSG_NAV_RELPOSNED** message. (Mostly common for MRO ZED-F9P modules because they arrive with connectors and wires).

This can be detected as the GPS heading is not working since **UBX_MSG_NAV_RELPOSNED** contains heading information.
If I2C wires are disconnected heading is working again.  

### Solution in this PR
This PR is sending commands to the GPS module to disable GPS protocols for I2C (UBX, NMEA, RTCM3X) at I2C if UART1 is selected as a physical communication protocol. After this is done GPS heading works again.


src/ubx.cpp
```diff
@@ -243,6 +243,23 @@ GPSDriverUBX::configure(unsigned &baudrate, const GPSConfig &config)
 			return -1;	// connection and/or baudrate detection failed
 		}
 
+		// Disable GPS protocols at I2C
```
bkueng, line 246: Can you do this in `configureDevice` which is used for the newer protocols?

Igor-Misic: Moved under `configureDevice`.

bkueng (approved):

PR #83, opened 2021-05-16
Remove UBX_CFG_KEY_CFG_I2COUTPROT_RTCM3X to fix M9N functionality

AlexKlimaj:
This PR removes the UBX_CFG_KEY_CFG_I2COUTPROT_RTCM3X message to fix U-blox M9N failing to start. Tested on a Pixhawk 4 with a M8N and M9N GPS. Also tested on an ARK CAN GPS with a M9N.


Igor-Misic:
@AlexKlimaj, thanks for testing and PR. 
Unfortunately, this will break the support for ZED-F9P where we need this to be disabled. Can you add something like this?
```
if (_board == Board::u_blox9_F9P) {
	cfgValset<uint8_t>(UBX_CFG_KEY_CFG_I2COUTPROT_RTCM3X, 0, cfg_valset_msg_size);
} 
```

AlexKlimaj:
Will do. 
An unrelated issue I just noticed on startup is this message.
` WARN  [gps] ubx msg 0x0104 invalid len 14930`

AlexKlimaj:
Looks like that works. You will have to test on the F9P before this can be merged.

bkueng (approved):

Igor-Misic (approved):

PR #84, opened 2021-08-02
Femto gps driver modify

Femtomes:
1. add UAVSTATUS message
2. Modify Femtomes GPS default message output frequency from 20Hz to 10Hz


src/femtomes.h
```diff
@@ -106,7 +132,7 @@ typedef union {
 * receive Femto complete uavgps frame
 */
 typedef struct {
-	uint8_t 		data[256];		/**< receive Frame message content */
+	uint8_t 		data[1024];		/**< receive Frame message content */
```
bkueng, line 135: femto_uav_status_t is 552 B in size, so you can reduce this to 600

src/femtomes.cpp
```diff
@@ -605,11 +617,32 @@ int GPSDriverFemto::configure(unsigned &baudrate, const GPSConfig &config)
...
 
 	if (_output_mode == OutputMode::GPS) {
+		if (writeAckedCommandFemto("LOG UAVGPSB 0.1\r\n", "<LOG OK", FEMTO_RESPONSE_TIMEOUT) == 0) {
+			FEMTO_DEBUG("Femto: command LOG UAVGPSB 0.1 success");
+
+		} else {
+			FEMTO_ERR("Femto: command LOG UAVGPSB 0.1 failed");
+		}
+
+		if (writeAckedCommandFemto("LOG UAVSTATUSB 0.1\r\n", "<LOG OK", FEMTO_RESPONSE_TIMEOUT) == 0) {
```
bkueng, line 627: The status could be enabled at a lower rate, to reduce link bandwidth. Probably 1 or 2 Hz is enough.
Also you could only enable it if `_satellite_info` is not null.

bkueng (approved):

bkueng:
Thanks for your updates.
I applied the code directly due to merge conflicts. Next time could you please start off latest master?

PR #87, opened 2021-08-17
Generic NMEA driver

bkueng:
From https://github.com/PX4/PX4-GPSDrivers/pull/74 with some cleanup & fixes.

Tested with a u-blox m8n set to NMEA output.

### DIFF
--- a/src/nmea.cpp
+++ b/src/nmea.cpp
@@ -1,0 +1,1034 @@
+/****************************************************************************
+ *
+ *   Copyright (c) 2020 PX4 Development Team. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ * 3. Neither the name PX4 nor the names of its contributors may be
+ *    used to endorse or promote products derived from this software
+ *    without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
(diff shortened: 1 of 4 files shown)


src/nmea.cpp
```diff
@@ -0,0 +1,1015 @@
...
+	int iRet = 0;
+
+	switch (_decode_state) {
+	/* First, look for sync1 */
+	case NMEADecodeState::uninit:
+		if (b == '$') {
+			_decode_state = NMEADecodeState::got_sync1;
+			_rx_buffer_bytes = 0;
+			_rx_buffer[_rx_buffer_bytes++] = b;
+		}
```
lukegluke, line 916:
```
} else if (b == RTCM3_PREAMBLE && _rtcm_parsing) {
	_decode_state = NMEADecodeState::decode_rtcm3;
	_rtcm_parsing->addByte(b);

}
```

@bkueng, probably `_rtcm_parsing` should be added same as in ashtech driver for PPK (0acaf6185245382389f7f6019cee6537d8f5a8e6) .

Kim-zhi-jiang:
@bkueng Positioned at first, move then circle around(in master version of yesterday)

UWB transmission to info mode(only 3):
$GPGGA,000522.953,2231.08747980,N,11354.04484670,E,1,04,0.030,-0.681,M,0.000,M,,0000*43
$GPGSA,A,3,0,1,2,3,,,,,,,,,2.550,0.030,2.549*39
$GPRMC,000522.953,A,2231.08747980,N,11354.04484670,E,000.007,125.498,080318,,,A*62

log: [here](https://logs.px4.io/plot_app?log=c825a49f-cc93-4f27-9515-de408631f425)

bkueng:
Thanks for testing @Kim-zhi-jiang, which exact GPS is that?

src/nmea.cpp
bkueng: Yes, can you send a PR?

Kim-zhi-jiang:
@bkueng [nooploop](https://www.nooploop.com/)
nnn

src/nmea.cpp
lukegluke: Done: https://github.com/PX4/PX4-GPSDrivers/pull/88 (but haven't tested it)

PR #88, opened 2021-08-18
nmea: add rtcm parsing and add debug ouput

lukegluke:
RTCM output in general receiver could be configured manually by user, so let support RTCM messages parsing just with warning.


bkueng (approved):
Thanks

PR #93, opened 2021-11-25
always using groundspeed for vel_m_s in ubx driver

sverling:
In the case where we subscribe to the UBX_MSG_NAV_VELNED message, the value we assign to vel_m_s is not the same as in the default case (UBX_MSG_NAV_PVT). Also according to the description of vel_m_s the value should correspond to the groundspeed (2d) and not the total 3d speed.


bkueng (approved):
Looks correct, thanks

Issue #96, opened 2022-02-14
Potential bug in GPSDriverUBX::receive 


chalkytoast:
Hello, I'm not sure if this is actually a bug, or the comments are inaccurate.

During configuration phase GPSDriverUBX::receive returns -1. This happens once, and when I ignore it, everything seems to work fine.

I have looked a bit into it, and the cause seems to be in reveive() when `handled` is set to -1 by GPSDriverUBX::parseChar, which according to the comments should return 0, 1 or 2. 

My workaround for now is to ignore the first -1 return, but I'd rather have a clean solution and appreciate any  clarification. Thanks.

bkueng:
Hi

Which GPS device is this? This could indeed be incorrect, can you trace it further?
Potentially here: https://github.com/PX4/PX4-GPSDrivers/blob/master/src/ubx.cpp#L1984

chalkytoast:
It's a Holybro H-RTK F9P. Your guess was right, that return appears to be the cause. It seems to happen after the survey-in is complete. Here is the stacktrace:

```
#0  GPSDriverUBX::payloadRxDone (this=0xb4703ec0) at src/driver/src/ubx.cpp:1984
#1  0x001182cc in GPSDriverUBX::parseChar (this=0xb4703ec0, b=213 '\325') at src/driver/src/ubx.cpp:1196
#2  0x001176cc in GPSDriverUBX::receive (this=0xb4703ec0, timeout=1300) at src/driver/src/ubx.cpp:1044
```

PR #97, opened 2022-02-17
fixes issue #96

chalkytoast:


bkueng (approved):
Thanks. Is RTK working for you even if `activateRTCMOutput` fails?

PR #99, opened 2022-03-22
Sbf heading

SeppeG:
Add sbf-heading


SeppeG:
The heading part of the code works, you can use both single and dual antenna receivers as well as a moving base setup. Outside of the [EKF2_AID_MASK] you don't have to change anything in the configuration or code to switch between single and dual antenna

GPSDriverSBF::configure --> This part of the code doesn't work, hence the block comment. I can' seem to get it working.

PR #100, opened 2022-03-22
Sbf heading

SeppeG:


SeppeG:
The heading part of the code works (additional test are needed), you can use both single and dual antenna receivers as well as a moving base setup. Outside of the [EKF2_AID_MASK] you don't have to change anything in the configuration or code to switch between single and dual antenna

GPSDriverSBF::configure --> This part of the code doesn't work, hence the block comment. I can' seem to get it working.

src/sbf.cpp
```diff
@@ -1,590 +1,629 @@
-/****************************************************************************
```
bkueng, line 1: Can you avoid changing file permissions and from what it looks like, line endings, and whitespace?

bkueng (commented):
> GPSDriverSBF::configure --> This part of the code doesn't work, hence the block comment. I can' seem to get it working.

What issue did you experience?

src/sbf.cpp
SeppeG: The commands get stuck in a loop, Mirko Denecke tested it as well and he got the same 

[TXT of loop](https://fromsmash.com/SBF-Heading-config-loop)

SeppeG:
Now the auto configuration is working as well. There was an issue with the setSBFOutput cmd in `sbf.h` (the cmd was too long).
Solution: 
```    "setSBFOutput, Stream1, COM1, DOP, msec100\n"\
    "setSBFOutput, Stream2, COM1, PVTGeodetic, msec100\n"\
    "setSBFOutput, Stream3, COM1, VelCovGeodetic, msec100\n"\
    "setSBFOutput, Stream4, COM1, AttEuler, msec100\n"\
    "setSBFOutput, Stream5, COM1, AttCovEuler, msec100\n"
```
Different streams with each 1 SBF block

bkueng:
@SeppeG can you run `make format` in px4 and commit the result?

SeppeG:
I added a pitch parameter and automatic com port detection (receiver side). 

I tested it with the mosaic-go dual and single antenna version.

src/sbf.cpp
```diff
@@ -1,590 +1,713 @@
...
+// -1 = error, 0 = no message handled, 1 = message handled, 2 = sat info message handled
+int GPSDriverSBF::receive(unsigned timeout)
+{
+	// Do not receive messages until we're configured
+	if (!_configured) {
+		gps_usleep(timeout * 1000);
+		return 0;
+	}
+
+	uint8_t buf[RECEIVE_BUF_SIZE];
```
bkueng, line 306: Why was this changed from GPS_READ_BUFFER_SIZE?

src/sbf.cpp
```diff
@@ -1,590 +1,713 @@
...
+		sendMessage(SBF_CONFIG_FORCE_INPUT);
+	}
+
+	// flush input and wait for at least 50 ms silence
+	decodeInit();
+	receive(50);
+	decodeInit();
+
+	char buf[GPS_READ_BUFFER_SIZE];
+	char com_port[5] = "";
```
bkueng, line 96:
```suggestion
	char com_port[5] {};

```

src/sbf.cpp
```diff
@@ -1,590 +1,713 @@
...
+
+	// Read buffer to get the COM port
+	do {
+		--offset; //overwrite the null-char
+		int ret = read(reinterpret_cast<uint8_t *>(buf) + offset, sizeof(buf) - offset - 1, SBF_CONFIG_TIMEOUT);
+
+		if (ret < 0) {
+			// something went wrong when polling or reading
+			SBF_WARN("sbf poll_or_read err");
+			return false;
```
bkueng, line 110:
```suggestion
			return ret;

```

src/sbf.cpp
```diff
@@ -1,590 +1,713 @@
...
+
+			int error_aux1 = _buf.payload_att_euler.error_aux1;
+			int error_aux2 = _buf.payload_att_euler.error_aux2;
+
+			// SBF_DEBUG("Mode: %u", _buf.payload_att_euler.mode)
+			if (error_aux1 == 0 && error_aux2 == 0) {
+				float heading = _buf.payload_att_euler.heading;
+				heading *= M_PI_F / 180.0f; // deg to rad, now in range [0, 2pi]
+
+				// heading -= _heading_offset; // range: [-pi, 3pi]
```
bkueng, line 614: Why is this commented?

bkueng (commented):
A couple of details, generally looks good.
The formatting is still off, did you run `make format`?

And can you provide a flight log?

src/sbf.cpp
SeppeG: The GPS_READ_BUFFER_SIZE is too big which causes the board to crash. I don't know if there is a fix for this, but the smaller size is enough and the program still works.

bkueng: `GPS_READ_BUFFER_SIZE` is smaller:
```
#define RECEIVE_BUF_SIZE 800
#define GPS_READ_BUFFER_SIZE 150
```

src/sbf.cpp
SeppeG: Because the heading offsets are calculated inside the receiver (so no need for that calculation), I send the setAttitudeOffset cmd at configuration to the receiver (line 175)

src/sbf.cpp
SeppeG: Yes, the GPS_READ_BUFFER_SIZE is large enough for the cmd replies but not enough for actual position data

bkueng: It does not have to be, it's assembled individually via `parseChar`

src/sbf.cpp
bkueng: Ok then better remove it here.

SeppeG:
```suggestion
```

src/sbf.cpp
SeppeG: True, i changed it to GPS_READ_BUFFER_SIZE

SeppeG: True, but if the buffer is not large enough, it will not always recognize an SBF block so the timeout will cause the receiver to be reinitialized.

src/sbf.cpp
```diff
@@ -37,33 +37,36 @@
...
+#include <cmath>
 
-#define SBF_CONFIG_TIMEOUT    500      // ms, timeout for waiting ACK
-#define SBF_PACKET_TIMEOUT    2        // ms, if now data during this delay assume that full update received
-#define DISABLE_MSG_INTERVAL  1000000  // us, try to disable message with this interval
+#define SBF_CONFIG_TIMEOUT        1000      // ms, timeout for waiting ACK
+#define SBF_PACKET_TIMEOUT        2        // ms, if now data during this delay assume that full update received
+#define DISABLE_MSG_INTERVAL    1000000  // us, try to disable message with this interval
+#define DNU                        100000.0 // Do-Not-Use value for PVTGeodetic
+#define MSG_SIZE                    100 // size of the message to be sent to the receiver.
```
SeppeG, line 56:
```suggestion
#define RECEIVE_BUF_SIZE        800
```

src/sbf.cpp
bkueng: Can you debug this, it does not make sense.
This looks incorrect: https://github.com/PX4/PX4-GPSDrivers/blob/ebaed6a08b8bdcf278144b4d79815115bbc8fae2/src/sbf.cpp#L463= (should be 0)
And the return value from payloadRxAdd should not be passed back to parseChar: https://github.com/PX4/PX4-GPSDrivers/blob/ebaed6a08b8bdcf278144b4d79815115bbc8fae2/src/sbf.cpp#L358=
https://github.com/PX4/PX4-GPSDrivers/blob/ebaed6a08b8bdcf278144b4d79815115bbc8fae2/src/sbf.cpp#L373=
You can compare with the ubx driver.

SeppeG: Never mind, it does work with GPS_READ_BUFFER_SIZE. I had to many output-streams on the receiver running

SeppeG:
I made the suggested changes and ran make format on the latest commit

bkueng (approved):
This looks good, I'm squashing the commits.

PR #101, opened 2022-03-25
ubx: add new sensor_gps_relative msg for NAV_RELPOSNED

dagar:


dagar:
Alternatively we could add a new GPSCallbackType to publish this and otherwise not have to pay for it when not used?

dagar:
Updated to add a new callback to simply publish the new relative message.

bkueng (approved):
What is the use-case, just logging?

PR #102, opened 2022-03-29
ubx: print relevent UBX-MON-VER output

dagar:
I was looking for an easy way to track specific firmware version of a ublox module. There are better ways to do this and capture the information properly per GPS driver instance PX4 side, but this looked like a potentially harmless start. Thoughts?

 - when decoding UBX-MON-VER print full module, firmware, protocol version output

``` Console
INFO  [gps] u-blox firmware version: SPG 3.01
INFO  [gps] u-blox protocol version: 18.00
INFO  [gps] u-blox module: NEO-M8N-0
```

### DIFF
--- a/src/ubx.cpp
+++ b/src/ubx.cpp
@@ -1720,13 +1720,33 @@ GPSDriverUBX::payloadRxAddMonVer(const uint8_t b)
 			// Part 2 complete: decode Part 2 buffer
 			UBX_DEBUG("VER ext \" %30s\"", _buf.payload_rx_mon_ver_part2.extension);
 
-			// in case of u-blox9 family, check if it's an F9P
-			if (_board == Board::u_blox9) {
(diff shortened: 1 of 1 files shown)


bkueng:
Good starting point, but can you print it on a single line?
We can also enable errors and warnings from the gps.

dagar:
> Good starting point, but can you print it on a single line? We can also enable errors and warnings from the gps.

The structure is a bit awkward with each line coming from a separate call to `payloadRxAddMonVer()`.

We could store these small strings, print a simple ublox summary after a successful configure, and maybe also expose it to the front end `gps status`.

bkueng (approved):
Ah right. Feel free to merge as is then.

dagar:
> Ah right. Feel free to merge as is then.

Thanks, I'll come back to this again. I also want to grab serial numbers per unit.

PR #104, opened 2022-04-12
Disable gps heading for UBX driver in RTK float fix type

JonasPerolini:
I've been testing the dual GPS heading using the ublox F9P chip. I get big angles spikes (yellow vertical lines in the attached file) as the gps fix_type changes from float RTK to fixed RTK. Furthermore, I see no changes in the heading accuracy. 

RTKFloatFixType

Finally, as expected, I've observed that the heading estimates are always more precise in RTK fixed: 

FixTypeResults

A simple solution, for now, is to disable the GPS heading when we are in RTK float fix type. 

Note: I believe that @dagar and @bresch have also experienced drops in accuracy when switching from RTK float to RTK fixed fix type.

### DIFF
--- a/src/ubx.cpp
+++ b/src/ubx.cpp
@@ -2039,13 +2039,7 @@ GPSDriverUBX::payloadRxDone()
 			bool carrier_solution_fixed = _buf.payload_rx_nav_relposned.flags & (1 << 4);
 			(void)rel_length_acc;
 
-			// RTK float fix type is not accurate enough 
-			if (!carrier_solution_fixed)
-			{
-				heading_valid = false; 
-			}
-
-			if (heading_valid && rel_pos_valid && rel_length < 1000.f) { // validity & sanity checks
+			if (heading_valid && rel_pos_valid && rel_length < 1000.f && carrier_solution_fixed) { // validity & sanity checks
 				heading *= M_PI_F / 180.0f; // deg to rad, now in range [0, 2pi]
 				heading -= _heading_offset; // range: [-pi, 3pi]
 
@@ -2103,11 +2097,6 @@ GPSDriverUBX::payloadRxDone()
 			gps_rel.reference_observations_miss  = _buf.payload_rx_nav_relposned.flags & (1 << 7);
 			gps_rel.heading_valid                = _buf.payload_rx_nav_relposned.flags & (1 << 8);
 			gps_rel.relative_position_normalized = _buf.payload_rx_nav_relposned.flags & (1 << 9);
-
-			if (!gps_rel.carrier_solution_fixed)
-			{
-				gps_rel.heading_valid = false; 
-			}
 
 			gotRelativePositionMessage(gps_rel);
 		}
--- a/src/ubx.cpp
+++ b/src/ubx.cpp
@@ -2043,8 +2043,6 @@ GPSDriverUBX::payloadRxDone()
 			if (_gps_position->fix_type != 6)
 			{
 				heading_valid = false; 
-				_gps_position->heading = NAN;
-				_gps_position->heading_acc = NAN; 
 			}
 
 			if (heading_valid && rel_pos_valid && rel_length < 1000.f) { // validity & sanity checks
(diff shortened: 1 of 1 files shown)


src/ubx.cpp
```diff
@@ -2038,6 +2039,14 @@ GPSDriverUBX::payloadRxDone()
 			bool rel_pos_valid = _buf.payload_rx_nav_relposned.flags & (1 << 2);
 			(void)rel_length_acc;
 
+			// RTK float fix type is not accurate enough 
+			if (_gps_position->fix_type != 6)
+			{
+				heading_valid = false; 
+				_gps_position->heading = NAN;
+				_gps_position->heading_acc = NAN;
```
bkueng, line 2047: Not needed
```suggestion
```

src/ubx.cpp
```diff
@@ -2076,8 +2085,17 @@ GPSDriverUBX::payloadRxDone()
...
+			if (_gps_position->fix_type != 6)
+			{
+				gps_rel.heading = NAN;
+				gps_rel.heading_acc = NAN; 
+			}
+			else
+			{
+				gps_rel.heading = _buf.payload_rx_nav_relposned.relPosHeading * 1e-5f * (M_PI_F / 180.f);  // 1e-5 deg -> radians
+				gps_rel.heading_accuracy = _buf.payload_rx_nav_relposned.accHeading * 1e-5f * (M_PI_F / 180.f); // 1e-5 deg -> radians
+			}
```
bkueng, line 2098: I wouldn't change this message, to still have the raw data.

bkueng (commented):
We can do this for now but I can imagine u-blox improves this with a firmware update or in the next hardware.
Which firmware version do you use?

JonasPerolini:
> We can do this for now but I can imagine u-blox improves this with a firmware update or in the next hardware. Which firmware version do you use?

HPG 1.30

JonasPerolini:
Thank you for the review @bkueng. I've pushed your suggestions.

bkueng (approved):
Thanks, looks good apart from the small style issue.

src/ubx.cpp
```diff
@@ -2038,6 +2038,12 @@ GPSDriverUBX::payloadRxDone()
 			bool rel_pos_valid = _buf.payload_rx_nav_relposned.flags & (1 << 2);
 			(void)rel_length_acc;
 
+			// RTK float fix type is not accurate enough
```
dagar, line 2041: Instead of depending on the fix_type being set I would directly check the `carrier_solution_floating` flag in NAV-RELPOSNED here.

See the carrier_solution_floating and carrier_solution_fixed handling below.

JonasPerolini: I've updated the code, please let me know if this is what you meant @dagar

dagar: I would roll carrier_solution_fixed right into the conditional below, but otherwise this looks good.

dagar (approved):

bkueng:
@JonasPerolini did you test the changes?

JonasPerolini:
**Overview:** 
1. As observed before, **RTK fixed** is excellent whereas **RTK float** is generally bad  
2. The heading accuracy is generally not representative of the quality of the heading 
3. Disabling the sensor_gps.00/heading in **RTK float** worked

**Setup:** 
- Two F9P GNSS units with software HPG 1.30 
- Baseline: 82.2 cm 

DualGNSS

GNSS

**Experiment:** 

**Benchmark (in Green)**: Move the rigid bar 3 times 45 deg back and forth in **RTK fixed** 

**Temper main GPS (in  blue)** Repeat the following 3 times: 

1.  Wait 20sec in RTK fixed
2. Place hand in front of main GPS antenna until we are in **RTK float**
4. Once in **RTK float**, move the rigid bar of 45 deg (clockwise or counterclockwise depending on the current position)
5. Wait 20 sec in **RTK float**
6. Remove hand to get into **RTK fixed** and 
7. Once in **RTK fixed** go back to 1. 

**Temper secondary GPS (in red):** same experiment as 2 but tempering the secondary GPS. 

**Results:** 
- Red curve: sensor_gnss_relative/heading 
- Blue curve: sensor_gps.00/heading 

1. As expected, in **RTK float** there are no sensor_gps.00/heading  values (see straight-line interpolation in the plot)
2. When in RTK fixed we have excellent results: around 0.1-degree absolute error. 
3. When passing from RTK float to RTK fixed, we sometimes observe jumps in the heading estimate (see purple vertical lines) 
4. When tempering with the second GPS, we sometimes go into DGPS fix type (so we lose heading). Once we recover the heading in RTK float, it is generally unusable. Once we pass back to RTK fixed, the heading is precise 

**Note:** 

1. The heading accuracy is usually not representative of the quality of the heading. For instance, during the 6.6-degree jump, the heading accuracy passes from 0.013 to 0.012. 
2. I've repeated the whole experiment 3 times and I obtain the same results.

PR #107, opened 2022-06-01
sbf: fix overrun on invalid length

### LINKED ISSUE #103: SBF driver
https://github.com/PX4/PX4-GPSDrivers/blob/ddb1825fe33f517853ca8a3ef75ac6f2df76f613/src/sbf.cpp#L393

If the length bytes are corrupt(_buf.Len is meaningless), the length arg to the crc16 call could overrun and cause hardfault. You can check it for %4 == 0 for sanity, and use _rxReadBytes instead of _buf.len to ensure you don’t go beyond sizeof(_buf)

fyi @skimslozo

julianoes:
Before checking the CRC across header and payload, we need to check if the length is valid. If the length is longer than what we have read into the buffer, it must be invalid and we don't have to bother with the CRC.

This should fix a potential segfault where the CRC overruns the allocated buffer due to a corrupt length field.

Fixes #103 as discovered by @devbharat.

### DIFF
--- a/src/sbf.cpp
+++ b/src/sbf.cpp
@@ -462,7 +462,9 @@ int GPSDriverSBF::payloadRxDone()
 	struct tm timeinfo;
 	time_t epoch;
 
-	if (_buf.length <= 4 || _buf.crc16 != crc16(reinterpret_cast<uint8_t *>(&_buf) + 4, _buf.length - 4)) {
+	if (_buf.length <= 4 ||
+		_buf.length > _rx_payload_index ||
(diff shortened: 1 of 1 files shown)


dagar:
@SeppeG

devbharat:
You can also use the fact that a valid length is always a multiple of 4.

bkueng (approved):

julianoes:
> You can also use the fact that a valid length is always a multiple of 4.

Right but it could potentially be too long and multiple of 4, right?

julianoes:
@devbharat oh you mean for optimization! I would think the packets usually don't get corrupt, and so the chances at all are very low, so therefore probably not worth the optimization.

devbharat (approved):

PR #112, opened 2022-08-31
ZED-F9P UART2 Baudrate Configurability and New UBX Mode.

stevesdawg:
## Describe the problem solved by this pull request
When using the U-blox ZED-F9P GPS module, users may want to use the GPS module's UART2 port to send or receive RTCM corrections using a Telemetry Radio, rather than hard wiring this port to another module. These radios operate at 57600 baud out of the box. But the UART2 baudrate is hard coded to 230400 baud. This PR allows us to configure the Baud Rate of the ZED-F9P's UART2 port through a QGC param.

Additionally, the existing modes that the **GPS_UBX_MODE** param covers misses one potential mode -- when a user is sending RTCM corrections from a _static_ base, or other _static_ correction source, directly to UART2. The existing modes that enable UART2 assume a moving base configuration. I've called this new mode **Rover With Static Base on UART2**. This mode is functionally equivalent to Rover with moving base on UART2 (Heading mode).

This PR to the GPS Driver must be merged alongside the corresponding PX4-Autopilot PR here: https://github.com/PX4/PX4-Autopilot/pull/20133

### DIFF
--- a/src/ubx.h
+++ b/src/ubx.h
@@ -980,7 +980,7 @@ private:
 	 * @param uart2_baudrate Baudrate of F9P's UART2 port
 	 * @return 0 on success, <0 on error
 	 */
-	int configureDevice(const GNSSSystemsMask &gnssSystems, int32_t uart2_baudrate);
+	int configureDevice(const GNSSSystemsMask &gnssSystems, const int32_t uart2_baudrate);
 	/**
 	 * Send configuration values and desired message rates (for protocol version < 27)
 	 * @param gnssSystems Set of GNSS systems to use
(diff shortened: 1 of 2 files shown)


src/ubx.h
```diff
@@ -1107,6 +1110,7 @@ class GPSDriverUBX : public GPSBaseStationSupport
 
 	const UBXMode _mode;
 	const float _heading_offset;
+	int32_t _uart2_baudrate;
```
bkueng, line 1113:
```suggestion
	const int32_t _uart2_baudrate;
```

stevesdawg: Thanks, added now

bkueng (approved):

PR #123, opened 2023-03-01
Add Unicore parser

julianoes:
This adds a separate class to parse Unicore messages which is then included as part of the NMEA parser.

This brings support for the Unicore UM982 module and heading specifically for the dual antenna setup.

The PR adds a unit test that can currently be run by a separate cmake command but can probably ultimately be included in other unit tests.

For now, the commands to run it, are:

```
cmake -Bbuild -H.
cmake --build build && build/gps-parser-test
```

FYI @vincentpoont2


src/crc.h
```diff
@@ -0,0 +1,38 @@
...
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ ****************************************************************************/
+
+#pragma once
+
+#include <cstdint>
+
+uint32_t calculateBlockCRC32(uint32_t length, uint8_t *buffer, uint32_t crc);
```
bkueng, line 38: Is there a more specific name you could use?

src/unicore.cpp
```diff
@@ -0,0 +1,149 @@
...
+		}
+
+		break;
+
+	case State::GotHashtag:
+		if (c == '*') {
+			_state = State::GotStar;
+
+			// Make sure buffer is zero terminated.
+			_buffer[_buffer_pos] = '\0';
```
bkueng, line 55: `_buffer_pos` can be equal to `sizeof(_buffer)` here.
You can change the check below to `if (_buffer_pos >= sizeof(_buffer)-1) {`

src/unicore.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+bool UnicoreParser::isHeading() const
+{
+	const char header[] = "UNIHEADINGA";
+
+	if (strncmp(header, _buffer, strlen(header)) != 0) {
+		return false;
+	}
+
+	return true;
```
bkueng, line 124: `return strncmp(header, _buffer, strlen(header)) == 0;`

src/unicore.cpp
```diff
@@ -0,0 +1,149 @@
...
+	if (ptr == NULL) {
+		return false;
+	}
+
+	ptr = strtok(ptr, ",");
+
+	unsigned i = 0;
+
+	while (ptr != NULL) {
+		ptr = strtok(NULL, ",");
```
bkueng, line 143: This needs another check for `ptr != NULL`

src/unicore.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+		} else if (i == 2) {
+			_heading.heading_deg = strtof(ptr, NULL);
+			break;
+		}
+
+		++i;
+	}
+
+	return true;
```
bkueng, line 158: `return i >= 5;`

bkueng (commented):
Looks good. Is only the heading a custom message, or does it also support a completely custom protocol? I'm just wondering if a completely separate driver would make sense.

> Stack is raised because we allocate 200 bytes for the Unicore parser.

The object is actually on the heap.

julianoes:
Thanks for the review @bkueng, as always spot on!

> Is only the heading a custom message, or does it also support a completely custom protocol? I'm just wondering if a completely separate driver would make sense.

The Unicore sentences are sprinkled in-between the NMEA messages, so it's still primarily NMEA. If it was a separate driver, we would have to duplicate the NMEA part, or extract NMEA parsing functions and re-use them. I would vote to do it the way it is now.

> The object is actually on the heap.

You are right, I'll revert that.

src/crc.h
julianoes: I just extracted it out of the femtomes driver. But you're right, I might as well document and name it properly.

src/unicore.cpp
julianoes: Oh, yes!

src/unicore.cpp
julianoes: That would be too nice :)

src/unicore.cpp
julianoes: Good catch!

src/unicore.cpp
julianoes: Actually, I was going to just early return when i == 5 was parsed.

bkueng (approved):
> The Unicore sentences are sprinkled in-between the NMEA messages, so it's still primarily NMEA. If it was a separate driver, we would have to duplicate the NMEA part, or extract NMEA parsing functions and re-use them. I would vote to do it the way it is now.

Alright, fine with me. It's well separated.

PR #128, opened 2023-04-03
Added support for ublox M10

katzfey:


src/ubx.cpp
```diff
@@ -829,8 +856,11 @@ bool GPSDriverUBX::cfgValsetPort(uint32_t key_id, uint8_t value, int &msg_size)
 			return false;
 		}
 
-		if (!cfgValset<uint8_t>(key_id + 3, value, msg_size)) {
-			return false;
+		// M10 has no UART2
```
bkueng, line 859:
```suggestion
		// M10 has no USB
```

src/ubx.cpp
```diff
@@ -140,36 +140,21 @@ GPSDriverUBX::configure(unsigned &baudrate, const GPSConfig &config)
...
 			cfgValset<uint8_t>(UBX_CFG_KEY_CFG_UART1OUTPROT_UBX, 1, cfg_valset_msg_size);
-
-			if (_output_mode != OutputMode::GPS) {
-				cfgValset<uint8_t>(UBX_CFG_KEY_CFG_UART1OUTPROT_RTCM3X, 1, cfg_valset_msg_size);
-			}
-
 			cfgValset<uint8_t>(UBX_CFG_KEY_CFG_UART1OUTPROT_NMEA, 0, cfg_valset_msg_size);
 			// TODO: are we ever connected to UART2?
 
-			// USB
```
bkueng, line 155: We need to check if connecting via USB still works

katzfey: I guess I don't understand. Those settings were not removed, just moved until after we know what kind of a ublox device were are connected to. If it is M10 then those values are not available.

src/ubx.cpp
katzfey: Yes, will change

src/ubx.cpp
bkueng: These settings were there because they might be needed for initialization (e.g. if ubx input is disabled). I know it's a chicken-egg problem at this point. So we need to test the changes.
Can you do so? You can use SITL and start the driver on your USB tty with a USB GPS attached.

katzfey: Unfortunately I don't have a way to test that. Another option is to just send those USB configuration keys before anything else and not require an ACK. M10 will just NACK that but it won't cause any harm.

bkueng: I tested it quickly with an F9P, it still works. Can you add a comment here that we configure USB below?

katzfey: Done!

bkueng (approved):

PR #129, opened 2023-05-02
Fixed compile errors for Qurt platform

katzfey:
Fixed some unused variable fatal errors and some incorrect pi constants so that the code compiles for Qurt platform.

### DIFF
--- a/src/sbf.cpp
+++ b/src/sbf.cpp
@@ -504,7 +504,7 @@ int GPSDriverSBF::payloadRxDone()
 		// Check boundaries and invalidate position
 		// We're not just checking for the do-not-use value (-2*10^10) but for any value beyond the specified max values
 		if (fabs(_buf.payload_pvt_geodetic.latitude) > (double) M_PI_2_F ||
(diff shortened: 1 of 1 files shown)


src/sbf.cpp
```diff
@@ -501,7 +503,8 @@ int GPSDriverSBF::payloadRxDone()
 
 		// Check boundaries and invalidate position
 		// We're not just checking for the do-not-use value (-2*10^10) but for any value beyond the specified max values
-		if (fabs(_buf.payload_pvt_geodetic.latitude) > M_PI_2 || fabs(_buf.payload_pvt_geodetic.longitude) > M_PI ||
+		if (fabs(_buf.payload_pvt_geodetic.latitude) > (double) M_PI_2_F ||
+		    fabs(_buf.payload_pvt_geodetic.longitude) > M_PI_PRECISE ||
```
dagar, line 507: Probably better to avoid px4 defines.h specifics like M_PI_PRECISE, this repo is used by other projects like QGC.

katzfey: Okay, went with M_PI_F then even though that is also defined in px4 defines.h

bkueng (approved):

PR #134, opened 2023-06-09
[BACKPORT v1.13] M10 support

julianoes:
This cherry-picks most of the UBX related commits to add M10 support for v1.13.


julianoes:
Let's use v1.14 instead.

PR #136, opened 2023-06-23
Adding MSG_NAV_HPPOSLLH for high precision RTK positioning + float64 lon/lat/alt

slgrobotics:
## Changes related to processing HPPOSLLH and converting lat/lon/alt values from int32 to float64

**Currently** LAT/LON/ALT values come mostly from UBX_MSG_NAV_PVT message, which does not deliver high precision components to *_gps_position* ORB message (a.k.a. SensorGps.msg).

**This PR seeks** to add parsing of the MSG_NAV_HPPOSLLH message and delivering high precision components of the coordinates to the PX4 codebase.

It also seeks to replace definitions of lat/lon/alt values in SensorGPS.msg from int32 to float64, which allows to preserve centimeter-level precision delivered by RTK GPS. These fields will be also renamed for clarity.

My code works fine on my rover, removing random jerking  which was caused by low precision "rounding" of the coordinates. With centimeter accuracy my rover (lawnmower) moves smoothly and precisely, which is critical for agricultural vehicles.

I have dual antenna F9P configuration and also take advantage of precise heading (RELPOSNED), as described here: https://github.com/PX4/PX4-GPSDrivers/pull/131

The values are also passed to EKF2 for better estimates and other consumers. That change is handled by the following PR:

https://github.com/PX4/PX4-Autopilot/pull/21678

**Note:**
This is how the mentioned values look like **PX4-Autopilot/msg/SensorGps.msg**, used to generate *sensor_gps_s* structure:

Now (after renaming):
```
...
float64 latitude_deg		# Latitude in degrees, allows centimeter level RTK precision
float64 longitude_deg		# Longitude in degrees, allows centimeter level RTK precision
float64 altitude_msl_m		# Altitude above MSL, meters
float64 altitude_ellipsoid_m	# Altitude above Ellipsoid, meters
...
```
Before:
```
int32 lat			# Latitude in 1E-7 degrees
int32 lon			# Longitude in 1E-7 degrees
int32 alt			# Altitude in 1E-3 meters above MSL, (millimetres)
int32 alt_ellipsoid 		# Altitude in 1E-3 meters bove Ellipsoid, (millimetres)
```


src/ubx.cpp
```diff
@@ -725,6 +725,8 @@ int GPSDriverUBX::configureDevice(const GPSConfig &config, const int32_t uart2_b
 		// heading output period 1 second
 		cfgValset<uint8_t>(UBX_CFG_KEY_MSGOUT_UBX_NAV_RELPOSNED_UART1, 1, cfg_valset_msg_size);
 		cfgValset<uint8_t>(UBX_CFG_KEY_MSGOUT_UBX_NAV_RELPOSNED_USB, 1, cfg_valset_msg_size);
+		cfgValset<uint8_t>(UBX_CFG_KEY_MSGOUT_UBX_NAV_HPPOSLLH_UART1, 1, cfg_valset_msg_size);
+		cfgValset<uint8_t>(UBX_CFG_KEY_MSGOUT_UBX_NAV_HPPOSLLH_USB, 1, cfg_valset_msg_size);
```
bkueng, line 729: Instead of all of these, can you instead add `cfgValsetPort(UBX_CFG_KEY_MSGOUT_UBX_NAV_HPPOSLLH_I2C, 1, cfg_valset_msg_size);` after https://github.com/PX4/PX4-GPSDrivers/blob/83e56c680a800847dc2b45098e4c220ae09f7497/src/ubx.cpp#L666

slgrobotics: Added definition of UBX_CFG_KEY_MSGOUT_UBX_NAV_HPPOSLLH_I2C, and the line (667) that allows that message on I2C.
We don't want to disallow HPPOSLLH on UARTs and USB, do we?

src/ubx.cpp
```diff
@@ -1882,13 +1901,17 @@ GPSDriverUBX::payloadRxDone()
...
-		_gps_position->alt_ellipsoid	= _buf.payload_rx_nav_pvt.height;
+		if (_gps_position->fix_type < 6) {
+			// When RTK is active and solid (fix=6), these values will be filled by HPPOSLLH:
+			_gps_position->latitude_deg		= _buf.payload_rx_nav_pvt.lat * 1e-7;
+			_gps_position->longitude_deg		= _buf.payload_rx_nav_pvt.lon * 1e-7;
+			_gps_position->altitude_msl_m		= _buf.payload_rx_nav_pvt.hMSL * 1e-3;
+			_gps_position->altitude_ellipsoid_m	= _buf.payload_rx_nav_pvt.height * 1e-3;
+
+			_gps_position->eph		= static_cast<float>(_buf.payload_rx_nav_pvt.hAcc) * 1e-3f;
+			_gps_position->epv		= static_cast<float>(_buf.payload_rx_nav_pvt.vAcc) * 1e-3f;
```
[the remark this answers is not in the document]
slgrobotics: Correct. I moved these two lines from ~1972 to 1915. It is conditional on actually updating the lat/lon/alt, which for PVT happens only when fix < 6.

src/ubx.cpp
slgrobotics: @bkueng  - thanks for explanation. I cleaned up **all** RELPOSNED and HPPOSLLH port setup, left only I2C as you suggested. Tested on my dual F9P RTK platform - everything works fine, these messages are coming via USB as before.

src/ubx.cpp
```diff
@@ -803,10 +800,6 @@ int GPSDriverUBX::configureDevice(const GPSConfig &config, const int32_t uart2_b
 		cfgValset<uint8_t>(UBX_CFG_KEY_CFG_UART1INPROT_NMEA, 0, cfg_valset_msg_size);
 		cfgValset<uint8_t>(UBX_CFG_KEY_CFG_UART1OUTPROT_UBX, 1, cfg_valset_msg_size);
 		cfgValset<uint8_t>(UBX_CFG_KEY_CFG_UART1OUTPROT_RTCM3X, 1, cfg_valset_msg_size);
-
-		cfgValset<uint8_t>(UBX_CFG_KEY_MSGOUT_UBX_NAV_RELPOSNED_UART1, 0, cfg_valset_msg_size);
```
bkueng, line 807: Just to double-check: as it's now enabled on both devices, are you getting this message only on 1 device (the rover)?

slgrobotics:
I will have to trace it if you really want exact answer, but my understanding is that only the main device will compute heading, and send it, while the secondary doesn't do it. I might be wrong. Is it important if it does?

Best Regards,
-- Sergei
________________________________

src/ubx.cpp
slgrobotics: I will have to trace it if you really want exact answer, but my understanding is that only the main ("rover") device will compute heading, and send it, while the secondary ("moving base") doesn't do it, as it only gets RTCM from the land base and doesn't have info about the second antenna location. I might be wrong. Is it important if messages are sent from both, as only the Main GPS is passing them to ORB?

slgrobotics: OK, I traced it and I think I am finally getting what you meant. The RELPOSNED was coming from both devices, but was ignored unless it was a "Rover with moving base" (ubx.cpp : 2155). So, I think it would be proper to only set it up under the same condition (my latest check-in ubx.cpp : 668).

For HPPOSLLH we better allow it for all cases, as it is beneficial to see actual coordinates of both antennas, using "gps status".

Sounds right?

bkueng: Ok thanks for checking. Yes that was my concern.

bkueng (approved):
Let's get this in

PR #137, opened 2023-07-26
Fixing UBlox M10 support

katzfey:
Some recent updates to UBX support broke M10. M10 has no RTCM so any attempts to set RTCM related configuration values will cause a configuration failure.


bkueng (approved):
Thanks

PR #140, opened 2023-08-28
unicore: parse AGRICA message, request heading

julianoes:
This add support to parse (but not use) the Unicore AGRICA message. This allows to determine whether we are talking to a UM982 and hence request the heading (UNIHEADINGA) message that we actually require.

FYI @vincentpoont2

### DIFF
--- a/src/nmea.cpp
+++ b/src/nmea.cpp
@@ -893,7 +893,7 @@ int GPSDriverNMEA::receive(unsigned timeout)
 
 				if (result == UnicoreParser::Result::GotHeading) {
 					++handled;
-					_unicore_heading_received_last = hrt_absolute_time();
+					_unicore_heading_received_last = gps_absolute_time();
 
 					// Unicore seems to publish heading and standard deviation of 0
 					// to signal that it has not initialized the heading yet.
@@ -920,7 +920,7 @@ int GPSDriverNMEA::receive(unsigned timeout)
 
 					// We don't use anything of that message at this point.
 
-					if (hrt_elapsed_time(&_unicore_heading_received_last) > 1000000) {
+					if (gps_absolute_time() - _unicore_heading_received_last > 1000000) {
 						request_unicore_heading_message();
 					}
 				}
@@ -961,8 +961,7 @@ void GPSDriverNMEA::request_unicore_heading_message()
 {
 	// Configure heading message on serial port at 5 Hz. Don't save it though.
 	uint8_t buf[] = "UNIHEADINGA COM1 0.2\r\n";
-	write(buf, sizeof(buf));
-
+	write(buf, sizeof(buf) - 1);
 }
 
 #define HEXDIGIT_CHAR(d) ((char)((d) + (((d) < 0xA) ? '0' : 'A'-0xA)))
--- a/src/nmea.cpp
+++ b/src/nmea.cpp
@@ -918,7 +918,9 @@ int GPSDriverNMEA::receive(unsigned timeout)
 				} else if (result == UnicoreParser::Result::GotAgrica) {
 					++handled;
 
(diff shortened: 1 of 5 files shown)


src/nmea.cpp
```diff
@@ -945,6 +957,14 @@ void GPSDriverNMEA::handleHeading(float heading_deg, float heading_stddev_deg)
 	_gps_position->heading_accuracy = heading_stddev_rad;
 }
 
+void GPSDriverNMEA::request_unicore_heading_message()
+{
+	// Configure heading message on serial port at 5 Hz. Don't save it though.
+	uint8_t buf[] = "UNIHEADINGA COM1 0.2\r\n";
+	write(buf, sizeof(buf));
```
bkueng, line 964: The terminating null should not be written I assume.
```suggestion
	write(buf, sizeof(buf)-1);
```

src/nmea.cpp
```diff
@@ -889,8 +889,11 @@ int GPSDriverNMEA::receive(unsigned timeout)
 					handled |= handleMessage(l);
 				}
 
-				if (_unicore_parser.parseChar(buf[i]) == UnicoreParser::Result::GotHeading) {
+				UnicoreParser::Result result = _unicore_parser.parseChar(buf[i]);
+
+				if (result == UnicoreParser::Result::GotHeading) {
 					++handled;
+					_unicore_heading_received_last = hrt_absolute_time();
```
bkueng, line 896:
```suggestion
					_unicore_heading_received_last = gps_absolute_time();
```

src/nmea.cpp
```diff
@@ -911,6 +914,15 @@ int GPSDriverNMEA::receive(unsigned timeout)
 						   (double)_unicore_parser.heading().heading_deg,
 						   (double)_unicore_parser.heading().heading_stddev_deg,
 						   (double)_unicore_parser.heading().baseline_m);
+
+				} else if (result == UnicoreParser::Result::GotAgrica) {
+					++handled;
+
+					// We don't use anything of that message at this point.
+
+					if (hrt_elapsed_time(&_unicore_heading_received_last) > 1000000) {
```
bkueng, line 923:
```suggestion
					if (gps_absolute_time() - _unicore_heading_received_last) > 1000000) {
```

src/nmea.h
```diff
@@ -71,8 +71,10 @@ class GPSDriverNMEA : public GPSHelper
 
 private:
 	void handleHeading(float heading_deg, float heading_stddev_deg);
+	void request_unicore_heading_message();
 
 	UnicoreParser _unicore_parser;
+	hrt_abstime _unicore_heading_received_last;
```
bkueng, line 77:
```suggestion
	gps_abstime _unicore_heading_received_last;
```

src/nmea.cpp
julianoes: Whoops, thanks.

src/nmea.cpp
julianoes: Right.

src/nmea.cpp
```diff
@@ -911,6 +914,15 @@ int GPSDriverNMEA::receive(unsigned timeout)
 						   (double)_unicore_parser.heading().heading_deg,
 						   (double)_unicore_parser.heading().heading_stddev_deg,
 						   (double)_unicore_parser.heading().baseline_m);
+
+				} else if (result == UnicoreParser::Result::GotAgrica) {
+					++handled;
```
bkueng, line 919: Could be worth adding your comment here to explain the logic:
`This allows to determine whether we are talking to a UM982 and hence request the heading (UNIHEADINGA) message that we actually require.`

bkueng (approved):

PR #143, opened 2023-11-21
ubx:Make it work with TX DMA

davids5:
The sendMessage we fragmenting messages by doing in multiple writes.
   in a DMA environment these writes might not be contiguous in time.

   The GPS would therefore see a runt message and not reply to it.


davids5:
Fixed in serial device driver

Issue #145, opened 2023-12-07
Wrong mapping when reading UBX-NAV-SAT messages


bedaberner:
It seems like "used" gets assigned the wrong value. 
https://github.com/PX4/PX4-GPSDrivers/blob/63990d218ec35ea965d634af6a79d3155561b743/src/ubx.cpp#L1723
it checks bit 1 of the flags when according to the [Interface Description](https://content.u-blox.com/sites/default/files/documents/u-blox-F9-HPG-1.32_InterfaceDescription_UBX-22008968.pdf) (page 150/151) it should be bit 3.

bkueng:
You are correct, can you create a pull request to fix it?

PR #146, opened 2023-12-11
ubx: fixed wrong mapping of UBX-NAV-SAT used parameter

bedaberner:
Fixed the wrong mapping described in  Issue https://github.com/PX4/PX4-GPSDrivers/issues/145.


bkueng (approved):
Thanks

PR #150, opened 2024-02-12
Ublox ZED-F9P-15B

AlexKlimaj:
This PR adds support for the Ublox ZED-F9P-15B.

It appears that moving base is not currently working. I think because it was NAKing `UBX_CFG_KEY_MSGOUT_RTCM_3X_TYPE4072_0_UART1` and I had to turn it off. 

It looks like it doesn't support moving base. I will need to update the driver to only allow moving base for the normal F9P.
https://content.u-blox.com/sites/default/files/ZED-F9P_ProductSummary_UBX-17005151.pdf

### DIFF
--- a/src/ubx.cpp
+++ b/src/ubx.cpp
@@ -562,7 +562,7 @@ int GPSDriverUBX::configureDevice(const GPSConfig &config, const int32_t uart2_b
 		rate_meas = 125; //8Hz for heading.
 
 	} else {
-		rate_meas = ((_board == Board::u_blox9_F9P) || (_board == Board::u_blox9_F9P_15B)) ? 100 : 125;
+		rate_meas = ((_board == Board::u_blox9_F9P_L1L2) || (_board == Board::u_blox9_F9P_L1L5)) ? 100 : 125;
 	}
 
 	cfgValset<uint16_t>(UBX_CFG_KEY_RATE_MEAS, rate_meas, cfg_valset_msg_size);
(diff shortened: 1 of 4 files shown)


src/ubx.cpp
```diff
@@ -1823,6 +1878,12 @@ GPSDriverUBX::payloadRxAddMonVer(const uint8_t b)
 
 			if (fwver_str != nullptr) {
 				GPS_INFO("u-blox firmware version: %s", fwver_str + strlen("FWVER="));
+
+				// Check if its a ZED-F9P-15B
+				if (strstr(fwver_str, "HPGL1L5")) {
```
dagar, line 1883: Is "HPGL1L5" going to be unique to this module or should we limit it to if hwVersion == "00190000" to be paranoid?

AlexKlimaj: See updated

src/ubx.cpp
```diff
@@ -1823,6 +1884,12 @@ GPSDriverUBX::payloadRxAddMonVer(const uint8_t b)
 
 			if (fwver_str != nullptr) {
 				GPS_INFO("u-blox firmware version: %s", fwver_str + strlen("FWVER="));
+
+				// Check if its a ZED-F9P-15B
+				if (strstr(fwver_str, "HPGL1L5") && (_board == Board::u_blox9)) {
```
dagar, line 1889: Nitpick, check the easy one first.

```suggestion
				if ((_board == Board::u_blox9) && strstr(fwver_str, "HPGL1L5")) {
```

src/ubx.h
```diff
@@ -1006,6 +1014,7 @@ class GPSDriverUBX : public GPSBaseStationSupport
 		u_blox9 = 9, ///< M9N, or any F9*, except F9P
 		u_blox9_F9P = 10, ///< F9P
 		u_blox10 = 11,
+		u_blox9_F9P_15B = 12, ///< ZED-F9P-15B
```
dagar, line 1017: These enums are ultimately for us, would something more descriptive like this work?

u_blox9_F9P -> u_blox9_F9P_L1L2
u_blox9_F9P_15B -> u_blox9_F9P_L1L5

julianoes (approved):

src/ubx.h
AlexKlimaj: Done

src/ubx.cpp
AlexKlimaj: Done

AlexKlimaj:
Rabsed on main and updated with dagars comments.

Not sure what we want to do with the moving base settings. The way it is right now, it will let you enable moving base on a L1L5 F9P, but you won't get a heading from the rover.

MaEtUgR:
Dev call: Moving base with the new chip doesn't work yet but better support it than not for this step.

PR #153, opened 2024-03-03
sbf: fix RTCM base functionality

julianoes:
This contains two fixes required to use the Septentrio Mosaic as an RTK base connected via USB to QGroundControl.

- Don't try to change baudrate of USB ports.
- Actually configure RTCM. This must have been lost somewhere in the past.

See commits for details.

Tested using the [Holybro H-RTK mosaic-H](https://holybro.com/products/h-rtk-mosaic)

FYI @SeptentrioGNSS and @vincentpoont2

### DIFF
--- a/src/sbf.cpp
+++ b/src/sbf.cpp
@@ -150,12 +150,14 @@ int GPSDriverSBF::configure(unsigned &baudrate, const GPSConfig &config)
 		return -1; // connection and/or baudrate detection failed
 	}
 
-	// Set baut rate
-	snprintf(msg, sizeof(msg), SBF_CONFIG_BAUDRATE, com_port, baudrate);
+	// Set baudrate, unless we're connected over USB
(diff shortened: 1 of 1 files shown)


AlexKlimaj (approved):

AlexKlimaj:
Looks good. It also appears that the driver doesn't support RTCM injection or output.

We can also update it to support moving base with two single antenna Mosaic's.

julianoes:
Oh, I haven't tested the other direction. Yes, I'll work on that then.

julianoes:
Let's merge this one for now, and then continue with more when I get to it.

julianoes:
I think it will configure it for RTCM, yes. Would you not want that? And is the config saved or only applied until reboot?

SeptentrioGNSS:
@julianoes , @AlexKlimaj another question, it seems like a fallback is in this code to GPS_CONFIG_RTCM. Who created this Septentrio config as I do not understand the combination into this one macro. Any feedback, otherwise we would like to fine tune this into a more generic config

PR #154, opened 2024-03-12
Add configuration for receiver logging

flyingthingsintothings:
This adds some new configuration that can be used to set up the receiver to log data to its internal disk. This is part of [a feature for PX4](https://github.com/PX4/PX4-Autopilot/pull/22874) to add parameters that enable logging setup by the user, without having to do it manually on the receiver. It is currently only implemented for the SBF driver. The implementation is kept generic to make sure other drivers can implement it as well.

### DIFF
--- a/src/gps_helper.h
+++ b/src/gps_helper.h
@@ -194,10 +194,27 @@ public:
 		I2C_OUT_PROT_RTCM3X = 1 << 5
 	};
 
+	enum class LoggingLevel : int8_t {
+		LITE = 0,
+		BASIC,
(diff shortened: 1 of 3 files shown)


flyingthingsintothings:
FYI @SeptentrioGNSS

SeptentrioGNSS:
Could this also include the CLI (setFileNaming, 'Cd', 'NamingType', 'filename')?

flyingthingsintothings:
Closing because this will be rebased onto the the new PX4 Septentrio driver.

PR #155, opened 2024-03-26
sbf: fix subsequent init in QGC

julianoes:
It needed a bit of trial and error to figure out why subsequent configurations in QGC would fail. I eventually settled on the current changes which basically:
- always set up input mode
- always disable all SBF messages to then enable them again
- remove all manual decoder inits and flushes

### DIFF
--- a/src/sbf.cpp
+++ b/src/sbf.cpp
@@ -1,6 +1,6 @@
 /****************************************************************************
  *
- *   Copyright (c) 2018 PX4 Development Team. All rights reserved.
+ *   Copyright (c) 2018-2024 PX4 Development Team. All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions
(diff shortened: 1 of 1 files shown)


julianoes:
@SeptentrioGNSS could you double check this makes sense please? Thanks!

julianoes:
Tested with QGC over USB and PX4 over serial.

SeptentrioGNSS:
> @SeptentrioGNSS could you double check this makes sense please? Thanks!

What should this fix? Not up to speed with the initial problem. Gives me a better perspective to check this.

julianoes:
I don't know really. What I do know is that it didn't work and kept going in a loop, and once I consolidated the logic a bit and removed the reads in-between the operations it started working. It was a bit of trial and error and the code paths for "normal" vs. "RTCM" were strangely different and intertwined making it hard to see what is going on.

julianoes:
Bump, if there are no objections I'm tempted to get this in soon.

PR #157, opened 2024-04-12
sbf: fix issue with automatic base config in QGC

flyingthingsintothings:
The automatic base configuration would sometimes print weird port names as well as reconnect every few seconds to the receiver.

### DIFF
--- a/src/sbf.cpp
+++ b/src/sbf.cpp
@@ -78,17 +78,29 @@ GPSDriverSBF::~GPSDriverSBF()
 
 int GPSDriverSBF::configure(unsigned &baudrate, const GPSConfig &config)
 {
+	char buf[GPS_READ_BUFFER_SIZE];
+	char msg[MSG_SIZE];
+
 	_configured = false;
 
 	setBaudrate(SBF_TX_CFG_PRT_BAUDRATE);
 	baudrate = SBF_TX_CFG_PRT_BAUDRATE;
 	_output_mode = config.output_mode;
 
+	// Make sure we can send commands to the receiver
 	sendMessage(SBF_CONFIG_FORCE_INPUT);
 
-	char buf[GPS_READ_BUFFER_SIZE];
-	char com_port[5] {};
+	// Disable previous output for now so we can detect the COM port
+	for (int i = 1; i <= 2; i++) {
+		snprintf(msg, sizeof(msg), SBF_CONFIG_DISABLE_OUTPUT, "COM", i);
+		sendMessageAndWaitForAck(msg, SBF_CONFIG_TIMEOUT);
+	}
+	for (int i = 1; i <= 4; i++) {
+		snprintf(msg, sizeof(msg), SBF_CONFIG_DISABLE_OUTPUT, "USB", i);
+		sendMessageAndWaitForAck(msg, SBF_CONFIG_TIMEOUT);
+	}
 
+	char com_port[5] {};
 	size_t offset = 1;
 	bool response_detected = false;
 	gps_abstime time_started = gps_absolute_time();
@@ -134,7 +146,6 @@ int GPSDriverSBF::configure(unsigned &baudrate, const GPSConfig &config)
 	}
 
 	// Delete all sbf outputs on current COM port to remove clutter data
-	char msg[MSG_SIZE];
 	snprintf(msg, sizeof(msg), SBF_CONFIG_RESET, com_port);
 
 	if (!sendMessageAndWaitForAck(msg, SBF_CONFIG_TIMEOUT)) {
@@ -185,12 +196,13 @@ int GPSDriverSBF::configure(unsigned &baudrate, const GPSConfig &config)
 		}
 
 		sendMessageAndWaitForAck(msg, SBF_CONFIG_TIMEOUT);
+		
+		snprintf(msg, sizeof(msg), SBF_CONFIG, com_port);
+
+		sendMessageAndWaitForAck(msg, SBF_CONFIG_TIMEOUT);
 	}
 
-	// Output a set of SBF blocks on a given connection at a regular interval.
 	int i = 0;
-	snprintf(msg, sizeof(msg), SBF_CONFIG, com_port);
-
 	do {
 		++i;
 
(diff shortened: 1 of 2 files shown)


flyingthingsintothings:
This builds on the PR by @julianoes that tried to fix automatic base setup of Septentrio receivers in QGroundControl. I've been noticing problems, even with his fixes applied. The problems are the following:

- The `configure()` function in `sbf.cpp` sometimes finds a weird port instead of an expected one like `USB1` or `COM1`. The port would be something like `$ùµé` or similarly broken. The reason for this is because lots of other information may be output by the receiver on the serial port, which is sometimes incorrectly interpreted as a serial port.
- The `parseChar()` function in `sbf.cpp` never returns a non-zero value when a complete RTCM message has been found. This lead the loop that processes incoming messages in QGroundControl to believe that the receiver wasn't set up correctly.
- SBF blocks were being output, even though they aren't used by the ground control station. Move all the SBF output setup behind a `if (!_output_type == Output::RTCM)` check so no SBF will be output to the serial port, causing problems for the `configure()` function and `parseChar()` function.

I'm working more on improving the automatic base support in QGroundControl and this submodule. This just fixes bugs that are already there today that prevent the setup from working correctly.

src/sbf.cpp
```diff
@@ -78,17 +78,29 @@ GPSDriverSBF::~GPSDriverSBF()
...
 	baudrate = SBF_TX_CFG_PRT_BAUDRATE;
 	_output_mode = config.output_mode;
 
+	// Make sure we can send commands to the receiver
 	sendMessage(SBF_CONFIG_FORCE_INPUT);
 
-	char buf[GPS_READ_BUFFER_SIZE];
-	char com_port[5] {};
+	// Disable previous output for now so we can detect the COM port
+	for (int i = 1; i <= 2; i++) {
```
julianoes, line 94: This had me very confused. I'm used to:
```suggestion
	for (int i = 0; i < 2; i++) {
```

src/sbf.cpp
```diff
@@ -282,21 +294,25 @@ bool GPSDriverSBF::sendMessageAndWaitForAck(const char *msg, const int timeout)
 	return found_response;
 }
 
-// -1 = error, 0 = no message handled, 1 = message handled, 2 = sat info message handled
+// return value:
+// 0b1111_1111 = an error occurred
+// 0b0000_0000 = no message handled (not set up yet)
+// 0b0000_0001 = message handled
+// 0b0000_0010 = sat info message handled
```
julianoes, line 301: If that's the return type I would make it an enum with these values, e.g.

```suggestion
enum class ReceiveResult : uint8_-t {
    Error = 0b11111111,
    NoMessageHandled = 0,
    MessageHandled = 0b01,
    SatInfoMessageHandled = 0b10,
};
```

That's if we can actually change the interface of the receive function.

And `0b11111111` doesn't work well for `int` type, as it's not -1 in that case.

src/sbf.h
```diff
@@ -65,19 +65,11 @@
...
-	"setSatelliteUsage, All\n" \
-	"setElevationMask, All, 10\n" \
-	"setReceiverDynamics, Moderate, Automotive\n" \
-	"setSBFOutput, Stream1, DSK1, Support, msec100\n" \
-	"setSBFOutput, Stream2, USB1, DOP+VelCovGeodetic, sec1\n" \
-	"setSBFOutput, Stream3, USB1, PVTGeodetic, msec200\n" \
-	"setFileNaming, DSK1, Incremental\n" \
-	"setFileNaming, DSK1, , 'px4rtcm'\n"
+	"setDataInOut, USB1, Auto, RTCMv3\n" \
+	"setPVTMode, Static, All, auto\n"
```
julianoes, line 72: :+1:

julianoes: Does this also include position information in case QGC wants to know where it is?

src/sbf.cpp
flyingthingsintothings: I always use the former for indices and the latter for real values (e.g. COM1 to COM2). I can change it if this is confusing.

src/sbf.h
flyingthingsintothings: I looked through the source code for any use of the SBF blocks but couldn't find any. How would I verify whether QGroundControl knows the base station position? Does it normally show up on the map? I tried with some different receivers as a base station but I don't see any indication of the base station position for any of them.

src/sbf.cpp
julianoes: I see. I guess it's fine.

src/sbf.h
julianoes: Oh ok, I thought that was how it worked but I might be dreaming :).

src/sbf.cpp
flyingthingsintothings: I spent some time looking at this but changing the return type of `receive()` alone doesn't seem like a good idea as it uses some other functions that themselves return `int` bitfields. It would add a lot of casts back and forth between int and custom return types. I'm particularly worried about changing `payloadRxDone()` as its implementation isn't documented that well so I don't know what it is supposed to return in which case.

`receive()` is also part of the `GPSHelper` class which is implemented by quite a few other classes. I think it might be better to leave this as is. I can remove the change to the documentation above the implementation of `receive()` as it's not completely correct (e.g. the `0b1111_1111`).

**All the signatures that would need updates, including to the parent class in the case of `receive()`:**
image

PS: I think updating this code/cleaning it up a bit is a good idea after a decision is made about <https://github.com/PX4/PX4-Autopilot/pull/22904> as lots of code can be removed or moved to QGroundControl if that one is merged.

julianoes: Oh I see, this is a general interface! Let's leave it as is then. Sorry, I thought that was new.

flyingthingsintothings: No problem. I'd be happy to have another look at it later and change the signature in `GPSHelper` if necessary, but for now I don't have the time to do these bigger refactoring tasks.

julianoes:
I haven't tested it myself but I'm fine with getting this in.

julianoes (approved):

PR #162, opened 2024-05-14
UBX UART TX packet consolidation

katzfey:
Added packet consolidation to UBX driver UART transmissions so that it sends each packet as a single transmission instead of as 3 separate ones (header, payload, and checksum). This is because on the VOXL 2 SLPI UART there is too much time between sequential transmissions that the GPS endpoint gives up and considers the packet lost before all portions can be sent.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,10 @@
+root = true
+
+[*]
+insert_final_newline = false
+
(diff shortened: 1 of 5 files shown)


katzfey:
I'm going to close this for now because there may be a clean way to resolve this in the VOXL 2 SLPI UART driver.

PR #165, opened 2024-06-29
sbf: Fix configuration messaging

niklaut:
If the GPS is already transmitting data messages (starting with `$@`), their payload contains `\0`, which makes strstr difficult to use since the search buffer "terminates" before a `$R` reponse is found.

- [x] Fix finding ACK when binary messages are received before.
- [x] Fix finding COM/USB port when binary messages are received before.
- [x] Fix finding COM/USB port when anything is received before (used wrong offset).
- [x] Fix reset commands (they where not detecting ACKs).
- [x] Fix detecting negative reponses `$R?` instead of `$R:`.
- [x] Fix not ignoring baudrate argument.
- [x] Fix configuration while GPS is already running.
- [x] Fix sending messages too quickly.
- [x] Integrate message formatting into sending functions to remove clutter.
- [x] Remove duplicate message sending due to non-functioning ACK finding.
- [x] Remove disabling of outputs, it's not necessary if ACK finding works
- [x] Add more (short) warnings to figure out what part of the config failed in the field.

Examples of hardware testing:

Cold boot vs hot boot.

Command - Binary message - Response

Hours of hot and cold configurations.

### DIFF
--- a/src/sbf.cpp
+++ b/src/sbf.cpp
@@ -37,6 +37,7 @@
  * Septentrio protocol as defined in PPSDK SBF Reference Guide 4.1.8
  *
  * @author Matej Frančeškin <Matej.Franceskin@gmail.com>
+ * @author Niklas Hauser <niklas@auterion.com>
  * @author <a href="https://github.com/SeppeG">Seppe Geuens</a>
  *
 */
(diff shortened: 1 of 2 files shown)


src/sbf.cpp
```diff
@@ -78,87 +79,40 @@ GPSDriverSBF::~GPSDriverSBF()
...
-
 	// Set baudrate, unless we're connected over USB
-	if (strncmp(com_port, "USB1", 4) != 0 && strncmp(com_port, "USB2", 4) != 0) {
-		snprintf(msg, sizeof(msg), SBF_CONFIG_BAUDRATE, com_port, baudrate);
-
-		if (!sendMessageAndWaitForAck(msg, SBF_CONFIG_TIMEOUT)) {
-			SBF_DEBUG("Connection and/or baudrate detection failed (SBF_CONFIG_BAUDRATE)");
+	if (strncmp(_com_port, "USB", 3) != 0) {
+		if (!sendMessageAndWaitForAck(SBF_CONFIG_BAUDRATE, _com_port, baudrate)) {
+			SBF_WARN("!BAUDRATE");
```
dagar, line 115: Keep the old descriptive message and/or move this back to DEBUG?

niklaut:
These fixes have been made obsolete by the new separate septentrio driver upstream

PR #168, opened 2024-08-21
Allow Custom Definitions Inclusion

HTRamsey:
@julianoes what about something simple like this? That way nothing changes for most people and those that want can define a new include


julianoes (approved):
Ah yes, that will do it!

PR #170, opened 2024-08-23
CI: Add gps-parser-test runner

HTRamsey:
@julianoes this will prevent the problem of making a mistake like I did


julianoes (approved):
Nice!

PR #171, opened 2024-08-26
unicore: extract vertical velocity and velocity accuracy

julianoes:
It looks like vertical velocity, and velocity accuracy is not available using NMEA, so we extract them from the Unicore ACRICA message instead.


src/unicore.cpp
```diff
@@ -179,5 +179,31 @@ bool UnicoreParser::extractHeading()
...
+	unsigned i = 0;
+
+	while (ptr != nullptr) {
+		ptr = strtok(nullptr, ",");
+
+		if (ptr == nullptr) {
+			return false;
+		}
+
+		if (i == 24) {
```
dagar, line 200: Where are you getting 24?

dagar: I'm not sure if I have the right document.

image

dagar: Including the speed accuracy is also super valuable.

julianoes: It's off by a few due to the header. Basically trial and error to get it right.

Ok, I will include speed accuracy.

julianoes: @dagar I've added speed accuracy, and moved all speed fields over to AGRICA.

julianoes:
Data from a flight test looks sane:

velocities
velocities and stddev

src/nmea.cpp
```diff
@@ -922,13 +926,24 @@ int GPSDriverNMEA::receive(unsigned timeout)
...
+
+					_gps_position->vel_m_s = _unicore_parser.agrica().velocity_m_s;
+					_gps_position->vel_n_m_s = _unicore_parser.agrica().velocity_north_m_s;
+					_gps_position->vel_e_m_s = _unicore_parser.agrica().velocity_east_m_s;
+					_gps_position->vel_d_m_s = -_unicore_parser.agrica().velocity_up_m_s;
+					_gps_position->s_variance_m_s =
+						(_unicore_parser.agrica().stddev_velocity_north_m_s * _unicore_parser.agrica().stddev_velocity_north_m_s +
+						 _unicore_parser.agrica().stddev_velocity_east_m_s * _unicore_parser.agrica().stddev_velocity_east_m_s +
+						 _unicore_parser.agrica().stddev_velocity_up_m_s * _unicore_parser.agrica().stddev_velocity_up_m_s)
+						/ 3.0f;
```
dagar, line 944: Shouldn't this be square root?

julianoes: Not sure, it's called variance...

PR #176, opened 2025-03-06
ubx: fix overflow

alexcekay:
While using a F9P with RTK, the parser suddenly got stuck in the RTCM parsing, due to `_decode_state` having an invalid state (out of the enum range). This causes a GPS driver restart, which causes a delay in receiving GPS data.

This was due to an overflow of the variable: `_buf` which is stored before `_decode_state` in memory:
```
	ubx_buf_t               _buf{};
	ubx_decode_state_t      _decode_state{};
```

Debugging showed, that an invalid length was parsed which is only partly checked in case of `UBX_MSG_RXM_RTCM`.
As that message has a fixed length, the easiest option is to directly check against the expected length.

### DIFF
--- a/src/ubx.cpp
+++ b/src/ubx.cpp
@@ -1528,7 +1528,7 @@ GPSDriverUBX::payloadRxInit()
 		break;
 
 	case UBX_MSG_RXM_RTCM:
-		if (_rx_payload_length < sizeof(ubx_payload_rx_rxm_rtcm_t)) {
+		if (_rx_payload_length != sizeof(ubx_payload_rx_rxm_rtcm_t)) {
 			_rx_state = UBX_RXMSG_ERROR_LENGTH;
 
 		} else if (!_configured) {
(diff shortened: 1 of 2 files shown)


AlexKlimaj (changes requested):
Is the RTCM message actually a fixed length? From my understanding, the RTCM messages are variable in length.

alexcekay:
Hi @AlexKlimaj,
RTCM messages themselves are variable length, that is correct. But this is a ublox monitor message, that is basically used to show some information about RTCM, not RTCM itself.

I found it first in u-blox8. Since then it was not changed from the fixed length of 8.
**u-blox 8:**
Screenshot from 2025-03-07 09-36-36

**u-blox 9:**
Screenshot from 2025-03-07 09-37-54

I also added the missing `msgType` to the parsed struct.

AlexKlimaj (approved):

PR #186, opened 2025-08-18
ubx: add new mode for GCS usage

alexcekay:
**Solved Problem**

In certain situations it is useful to have a GPS contained in a ground control station. Other devices in the ground station need to be able to access the GPS data.

**Solution**

 To allow other devices in the ground station to access the GPS data, a new mode is introduced that configures UBX GPS to output NMEA data on UART2. Then devices can connect to UART2 of the GPS module and have access to NMEA data.

### DIFF
--- a/src/ubx.cpp
+++ b/src/ubx.cpp
@@ -939,6 +939,27 @@ int GPSDriverUBX::configureDevice(const GPSConfig &config, const int32_t uart2_b
 			return -1;
 		}
 
+	} else if (_mode == UBXMode::GroundControlStation) {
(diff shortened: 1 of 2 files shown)


alexcekay:
Hi @dakejahl, basically yes. The name basically reflects the way we currently use it (in our GCS), but maybe we can find a more generic name. What comes to my mind is something like:
- UART2Output
- ExternalReceiver
- AdditionalOutput

niklaut (approved):
I also cannot come up with a better name…

PR #189, opened 2025-10-07
Expose u-Blox min elevation, min SNR and DGNSS timeout for RTK Fix Improvement

wrsolichin:
This PR increases the minimum elevation for u-Blox GPS from 10 (default) to 15, which is useful for RTK Fix improvement as discussed in the following articles: https://mowerproject.com/2020/07/06/an-even-better-rtk-fix/ and https://discuss.cubepilot.org/t/here-does-not-hold-rtk-fixed-status-reliably-help-please/1056/64. The configuration item address is defined as below.

Screenshot from 2025-10-07 11-30-59

Currently, I have only set the hard coded value for the u-Blox GPS for the minimum elevation, but if it is useful we can expose it to PX4 autopilot, and also optionally expose the `CFG-NAVSPG-INFIL_MINCNO` for further improvement as discussed in the article.

### DIFF
--- a/src/ubx.cpp
+++ b/src/ubx.cpp
@@ -75,6 +75,9 @@ GPSDriverUBX::GPSDriverUBX(Interface gpsInterface, GPSCallbackPtr callback, void
 	_gps_position(gps_position),
 	_satellite_info(satellite_info),
 	_dyn_model(settings.dynamic_model),
+	_min_cno(settings.min_cno),
+	_min_elev(settings.min_elev),
+	_dgnss_timeout(settings.dgnss_timeout),
 	_mode(settings.mode),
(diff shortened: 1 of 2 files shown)


dakejahl:
Thanks for this! We should definitely make this configurable with PX4 parameters. Can you make the minimum elevation a member variable and add it the constructor? I'd also add the SNR while we're here and thinking about it. For now let's leave them at their previous default and once we do some testing with the new values (elevation 15, SNR 35) we can decide on if it's an overall improvement.

dakejahl:
Any testing updates to share? Improvements with RTK lock?

dakejahl (approved):

PR #191, opened 2025-11-14
ubx: fix M9N output rate to increase sat count

AlexKlimaj:
Needs testing. Fixes https://discuss.px4.io/t/m9n-gps-is-limited-to-16-sats-with-1-16/47732


dakejahl (approved):

dakejahl:
> satellites_used: 8

Did you check it gets more than 16?

PR #201, opened 2026-03-04
fix(ubx): increase RTCM 1005 output rate for moving base

dakejahl:
Change RTCM 3.x Type 1005 (base station ARP) output rate from every 5th measurement to every measurement. With tighter DGNSS timeouts, the previous 5s interval can cause intermittent RTK fix loss on the rover because the base station position message arrived too infrequently.

Supersedes #109


AlexKlimaj (approved):

mrpollo:
What's missing to get this merged?

PR #216, opened 2026-07-27
feat(ubx): decode NAV-DAHEADING for X20 dual antenna heading

isak-aslund:
### Summary

Adds dual antenna heading support for the u-blox X20 platform (ZED-X20D).

These receivers report the antenna 1 → antenna 2 baseline in **UBX-NAV-DAHEADING** (0x01 0x45), a message the driver did not know. UBX-NAV-RELPOSNED still carries the baseline to the *RTK reference station* on the X20, so it is not a substitute — a ZED-X20D currently has no heading source at all and `sensor_gps.heading` stays `NAN`.

The message is decoded into `sensor_gps.heading` and `sensor_gnss_relative`.

Note the flag layout differs from NAV-RELPOSNED: `relPosHeadingValid` is **bit 6** (not bit 8), and the vector is plain millimetres with no high-precision components.

### Choosing the heading source

NAV-DAHEADING is not requested in `RoverWithMovingBaseUART1/2`. Those are the only modes where NAV-RELPOSNED is also enabled, and there the two messages report different baselines (moving base → rover vs. the receiver's own ANT1 → ANT2) while `GPS_YAW_OFFSET` is a single value describing only one of them. Deciding at configure time keeps the two mutually exclusive, instead of letting whichever message arrives first take over at runtime.

### Known behaviour: occasional missing sample

`sensor_gps.heading` is cleared after every publication, and NAV-DAHEADING is not guaranteed to be decoded between two NAV-PVT messages — it sometimes lands just *after* the one that triggered the publish, so that sample goes out with no heading.

Measured on a ZED-X20D at 10 Hz over 3239 publications, counting NAV-DAHEADING messages per NAV-PVT-to-NAV-PVT window:

| msgs in window | windows | heading NaN |
|---|---|---|
| 0 | 130 | 130 |
| 1 | 2979 | 0 |
| 2 | 130 | 0 |

130 and 130, exactly balanced — nothing is lost, the message slips by one window. EKF2 tolerates this (`reset_timeout_max` is 7 s), so no compensation is done here. The visible symptom, `GPS_RAW_INT` reporting yaw 0, is a separate MAVLink encoding bug fixed in PX4/PX4-Autopilot#28095. PX4/PX4-Autopilot#27102 removes the coupling entirely by giving heading its own topic and timestamp — this change produces the `sensor_gnss_relative` stream that consumes.

### Compatibility

The config key is sent in its own CFG-VALSET because it does not exist on position-only X20 firmware, where a NAK must not abort the rest of the configuration. Everything else is gated on `Board::u_blox_X20`, so no other receiver is affected.

### Testing

- **ZED-X20D** (HDG 2.00, UART @ 115200, 10 Hz, CubeOrange Plus): heading validated against a tape-measured baseline at 0.62 m and 1.5 m — reported `position_length` matched to a few cm, heading matched mag based heading (RM3100), and `heading_accuracy` scaled with baseline length as expected.
- **ZED-X20P**: configures and reports position normally, confirming the NAK path.
- **Not tested**: SPI/I2C, DroneCAN, nav rates other than 10 Hz, EKF2 yaw fusion (`EKF2_GPS_CTRL` bit 3).


isak-aslund:
This is my first PX4 PR, feedback is appriciated. 
@dakejahl has been patching the X20 driver and could perhaps provide a review?

dakejahl:
**Claude review on behalf of @dakejahl**

**1. Remove the heading carry-over** (`src/ubx.cpp:2391-2398`, `2717-2720`)

`receive()` returns as soon as NAV-PVT arrives (`src/ubx.cpp:1470`), so there is one publication per PVT. Since PVT precedes DAHEADING on the wire, `_daheading_valid` set in window N is consumed by the PVT of window N+1: in a window that also carries a fresh DAHEADING the carried value is overwritten before publish (no effect), and in a "0 messages" window it publishes the previous epoch's heading with this epoch's timestamp and the original `accHeading`.

That is 100 ms of unmodelled latency on ~4% of samples — `yaw_rate × 0.1 s`, so 6° at the `MPC_YAWRAUTO_MAX` default of 60 deg/s, handed to EKF2 at ~1° stated confidence. Either it is fused as a wrong yaw or `EKF2_HDG_GATE` throws it out, and thrown out is exactly the NaN you were avoiding. The correct sample is not lost either: DAHEADING_{N+1} is decoded at the start of window N+2 and then overwritten by DAHEADING_{N+2}. The driver discards a fresh heading to publish an older one.

Any correct version of this is a no-op — a heading may only be attached to the PVT of its own epoch, and if it were attached there would be nothing to carry. If you want no gaps, align the epoch instead: both messages carry `iTOW`, so hold `receive()` after NAV-PVT until the matching NAV-DAHEADING is decoded, bounded by `UBX_PACKET_TIMEOUT` and with a fallback, since the outer timeout returns -1 and restarts the driver.

**2. The `GPS_RAW_INT` symptom is a PX4-side encoding bug**

`GPS_RAW_INT.hpp:108-119` leaves `msg.yaw` at 0 when `sensor_gps.heading` is NaN. Per the message spec, 0 means "this GPS does not provide yaw"; a yaw-capable receiver that momentarily has none should send `UINT16_MAX`. Worth fixing there — it is a real protocol violation, and it removes the motivation for the driver-side cache.

**3. Precedence switch is a runtime race with a shared offset** (`src/ubx.cpp:2682`, `2629`)

NAV-RELPOSNED is only enabled in `RoverWithMovingBaseUART1/2` (`src/ubx.cpp:943`), the one configuration where both sources exist — and there they are different baselines (moving base → rover antenna vs. the X20D's own antenna 1 → antenna 2) while `GPS_YAW_OFFSET` is a single value applied to both (`src/ubx.cpp:2579`, `2705`). Decide it at configure time: skip the VALSET at `src/ubx.cpp:969` in moving-base rover modes.

**Nits**
- `_daheading_valid` is not reset alongside `_use_nav_daheading = false` (`src/ubx.cpp:968`), so a heading from before a re-`configure()` attaches to the first PVT after it. Moot if the cache goes.
- `src/ubx.cpp:2696-2712` duplicates `2574-2589` verbatim; a small helper would keep the two from diverging.

isak-aslund:
Thanks for the review, all taken.

**1.** Carry-over removed. You're right that any correct version should be a no-op, and the failure mode is worse than the gap it hides: an old heading under a new timestamp with the original `accHeading` either gets fused as a wrong yaw or thrown out by `EKF2_HDG_GATE`. Skipped the iTOW alignment — `reset_timeout_max` is 7 s, so EKF2 tolerates a missing sample and that doesn't justify blocking `receive()`.

**2.** Didn't know the spec said UINT16_MAX for a heading outputting receiver not being able to produce heading. This is a core PX4 change, I drafted up a change, please see PX4/PX4-Autopilot#28095.

**3.** Now decided at configure time — NAV-DAHEADING isn't requested in `RoverWithMovingBaseUART1/2`. Since NAV-RELPOSNED is only enabled in exactly those two modes, the two sources are now mutually exclusive by configuration, so `_use_nav_daheading` and its guard are gone entirely and the NAV-RELPOSNED path is back to its original behaviour.

**Nits:** the un-reset flag went with the cache. Heading conversion is now a shared `relPosHeadingToYaw()` helper used by both handlers.
**Edit:** PR body updated to reflect the changes after review.

## Extra info
Worth flagging PX4/PX4-Autopilot#27102 for anyone following this: under that, heading reaches EKF2 through `sensor_gnss_relative` with its own timestamp and buffer, with no NAV-PVT pairing at all. On this receiver (X20D) `sensor_gnss_relative` logged 3476 samples against 3475 `sensor_gps`, and `heading_valid` was never false after convergence — so that path never had the coupling this PR was working around. This change produces exactly that stream, so the two compose rather than overlap.

isak-aslund:
Performed [flight test](https://review.px4.io/plot_app?log=3cabbcfc-7578-44af-99c1-e6b9b3cc34d9) today with custom 30" Xframe, 0.6m separated antennas. Straight, turns, yaw spins, position mode and altitude mode.

Wow, this receiver is a beast!
Reported horizontal and vertical accuracy 0.5m and 0.8m for entire flight. Consistently 32 sats.
Heading was ~0.8deg accuracy spiking to 0.95deg during harder altitude turns/stops.
Heading was valid for 1142/1145 (99.7%), EKF2 fused all available samples.

I'm happy with the performance and working implementation as is now. 
Anything more you'd like to reivew / change @dakejahl ?

isak-aslund:
Follow-up commit after some CLAUDE reviews

  - **Always write the NAV-DAHEADING rate** instead of skipping it. The rate lives in the receiver's RAM config, so skipping left it enabled across a mode change — NAV-DAHEADING would then stream alongside NAV-RELPOSNED, win the epoch, and be interpreted with the wrong `GPS_YAW_OFFSET`. Writing it unconditionally on X20 makes a mode change idempotent; a NAK from a position-only X20P is still tolerated.
  - **Also exclude the moving base half** of the pair via `isMovingBase()`, not just the rover half.
  - Folded the 10 m baseline limit into a named constant — it was duplicated as `1000.f` cm and `10.f` m.

dakejahl (approved):
You don't have a dual GPS for MovingBase+Rover Heading do you? I'll have to test that to double check.

isak-aslund:
> You don't have a dual GPS for MovingBase+Rover Heading do you? I'll have to test that to double check.

I got 2x X20P modules, I'll test MovingBase+Rover heading with them to make sure.

isak-aslund:
Did some testing and have now updated with the following:

## 1 `DAHEADING_OFFSET`
In the module there is a `CFG-NAVSPG-DAHEADING_OFFSET` which practically does the same things as PX4 `GPS_YAW_OFFSET`.
If for some reason this would be configured on the module itself in flash it would conflict with GPS_YAW_OFFSET, I verified by setting it to 45deg and getting 45deg error of heading in PX4.
The driver now writes  `CFG-NAVSPG-DAHEADING_OFFSET = 0` to enforce that GPS_YAW_OFFSET will always be correct.
Tested on hardware by manually setting   `CFG-NAVSPG-DAHEADING_OFFSET = 4500` (45deg) and then rebooting PX4 (without power cycling gps module) and observing it gets set to 0 after PX4 boot.

## 2 Moving base + Rover on 2xX20P
I gave moving baseline a try with 2x X20P modules and it did not work with `GPS_UBX_MODE= 3` (passthrough). 

The correction message RTCM 4072.0 was never enabled which is required. Only F9P had that enabled.  
The module now sets that message up correctly. 
I've verified on bench that both modes (`GPS_UBX_MODE= 1 and 3` work now and I get heading. 

## 3 Moving base + Rover with 1xX20P and 1xX20D
Quick test with `GPS_UBX_MODE = 3`, X20P as rover and X20D as moving base.
Rover reports heading after ~1min. The X20D publishes none of its own, which is what `isMovingBase()` is for - it would otherwise report a second heading from its own antenna pair using the same `GPS_YAW_OFFSET`.

I don't have the possibility to do `GPS_UBX_MODE = 1` with UART2 on this setup.


@dakejahl Feel free to merge.

dakejahl:
@isak-aslund thanks for the thorough testing!

PR #200, opened 2026-02-09
ubx: diff corrections age, fix quality, agc, antenna status and power

dakejahl:
## Summary

Surface additional u-blox receiver state via `sensor_gps`:

- **`diff_age`** — differential correction age in seconds, from `UBX-NAV-PVT.flags3.lastCorrectionAge` via a lookup table (binned 0-12 → 0-120 s).
- **`automatic_gain_control`** — from `UBX-MON-RF.agcCnt` (0-8191, % of max gain).
- **`antenna_status`, `antenna_power`** — from `UBX-MON-RF.antStatus` / `antPower`.
- **`SYSTEM_ERROR_ANTENNA`** — set when `antStatus` reports SHORT or OPEN, cleared otherwise.

Also extends `ubx_payload_rx_nav_pvt_t` (adds `flags3`, `magDec`, `magAcc`) to match the ubx8+ spec and make `flags3` addressable.

## Not populated

Abstract quality metrics (`corrections_quality`, `system_status_summary`, `gnss_signal_quality`, `post_processing_quality` on `sensor_gnss_status`) are left at 255 — u-blox has no 0-10 metric equivalent to Septentrio's `QualityInd`. Synthesizing from `fix_type` would be circular.

## Related

- PX4-Autopilot integration: PX4/PX4-Autopilot#26438
- DSDL `Quality` message: dronecan/DSDL#77

PR #218, opened 2026-08-04
ubx: log automatic_gain_control correctly

mbjd:
`automatic_gain_control` was already filled from `agcCnt` in the `MON-HW` paths:

https://github.com/PX4/PX4-GPSDrivers/commit/8ba8d9f19c2328955424d035d6ce5a08ef11a371

But here it is also available [1], so let's populate the topic.

1: https://content.u-blox.com/sites/default/files/documents/u-blox-F9-HPG-1.32_InterfaceDescription_UBX-22008968.pdf, Section 3.14.9, p133


dakejahl (approved):
There's some other stuff in here you might be interested in
https://github.com/PX4/PX4-GPSDrivers/pull/200/changes

PR #222, opened 2026-08-06
feat(ubx): add support for UBX_MSG_MON_SPAN and full support of UBX_MON_RF

ThomasRigi:
Updated version of the submodule for supporting UBX_MSG_MON_SPAN

See https://github.com/PX4/PX4-Autopilot/pull/20474 for the initial version by @dagar .

Main PR: https://github.com/PX4/PX4-Autopilot/pull/28173


src/ubx.cpp
```diff
@@ -1239,6 +1202,22 @@ int GPSDriverUBX::configureDevice(const GPSConfig &config, const int32_t uart2_b
 		}
 	}
 
+	// UBX_MSG_MON_SPAN
+	cfg_valset_msg_size = initCfgValset();
+	uint8_t value = _spectrum_analyzer ? 1 : 0;
+	cfgValset<uint8_t>(UBX_CFG_KEY_MSGOUT_UBX_MON_SPAN_UART1, value, cfg_valset_msg_size);
```
ThomasRigi, line 1208: should we pass through cfgValsetPort with the _I2C key ? I don't see the point of configuring the USB port though, so I prefer the direct UART1 config we have here.

ThomasRigi: also, is there the possibility that this key doesn't exist on any protocol 27+ GNSS ? We only have F9P to test, and there it works.

src/ubx.cpp
```diff
@@ -2789,13 +2709,45 @@ GPSDriverUBX::payloadRxDone()
...
 		_gps_position->jamming_state		= _buf.payload_rx_mon_rf.block[0].flags;
 
 		ret = 1;
 		break;
 
+	case UBX_MSG_MON_SPAN:
+		UBX_TRACE_RXMSG("Rx MON-SPAN");
+
+		if (_spectrum_analyzer) {
+			const uint64_t timestamp_sample = gps_absolute_time(); // TODO: adjust with delay estimate
```
ThomasRigi, line 2756: TODO ported from #20474. Do we really need timestamp_sample in the message ? I don't think it's such a time-critical information.

src/ubx.cpp
```diff
@@ -1239,6 +1202,22 @@ int GPSDriverUBX::configureDevice(const GPSConfig &config, const int32_t uart2_b
...
+
+	UBX_DEBUG("Configuration spectrum analyzer");
+	if (!sendMessage(UBX_MSG_CFG_VALSET, _tx_cfg_valset_buf, cfg_valset_msg_size)) {
+		UBX_WARN("configuring UBX_MSG_MON_SPAN FAILED");
+		return -1;
+	}
+
+	if (waitForAck(UBX_MSG_CFG_VALSET, UBX_CONFIG_TIMEOUT, true) < 0) {
+		fprintf(stderr, "configuring UBX_MSG_MON_SPAN FAILED");
+		return -1;
```
dakejahl, line 1219: this should fail gracefully not return -1, for receivers/fw that don't support the key

src/ubx.cpp
```diff
@@ -1879,6 +1847,19 @@ GPSDriverUBX::payloadRxInit()
 
 		break;
 
+	case UBX_MSG_MON_SPAN:
+		if ((_rx_payload_length > sizeof(ubx_payload_rx_mon_span_t)) ||
+		    (_rx_payload_length == 4) ||
+		    (_rx_payload_length - 4) % sizeof(ubx_payload_rx_mon_span_t::ubx_payload_rx_mon_span_block_t) != 0) {
```
dakejahl: magic 4?

src/ubx.cpp
dakejahl: I think it doesn't hurt, to time align with the PVT solution

src/ubx.cpp
```diff
@@ -2787,11 +2708,72 @@ GPSDriverUBX::payloadRxDone()
 
 	case UBX_MSG_MON_RF:
 		UBX_TRACE_RXMSG("Rx MON-RF");
+		{
+			// Fill fields for SensorGps
+			// todo: remove them from the message definition, as they're now logged twice ?
+			_gps_position->noise_per_ms		= _buf.payload_rx_mon_rf.block[0].noisePerMS;
+			_gps_position->automatic_gain_control	= _buf.payload_rx_mon_rf.block[0].agcCnt;
+			_gps_position->jamming_indicator	= _buf.payload_rx_mon_rf.block[0].jamInd;
+			_gps_position->jamming_state		= _buf.payload_rx_mon_rf.block[0].flags;
```
ThomasRigi, line 2718: Porting from https://github.com/PX4/PX4-GPSDrivers/pull/223 :

These fields are now logged in a duplicate for ubx receivers.

In our fork we removed them, but that would cause issues in general because we're removing info from drivers .

automatic_gain_control is used by no other driver, so we could definitely remove that one
noise_per_ms is used by femtomes
jamming_indicator used by multiple
jamming_state also used by multiple

src/ubx.cpp
ThomasRigi: I was under the impression that all ublox gnss that are on protocol version 27 or newer do support the key. I asked chatGPT for an overview and then verified all the modules that it does list as 27+, and they all have the key. And for older modules I couldn't find any document with 27+. For example, here the most recent document from the ublox website on M8 : https://content.u-blox.com/sites/default/files/products/documents/u-blox8-M8_ReceiverDescrProtSpec_UBX-13003221.pdf : it only goes to protocol version 23.01.

So in conclusion I don't think that there's any real reason for this configuration to fail, but as said above I don't have all the hardware to explicitly test it. 

That said, I'm also ok with just removing the `return -1` from both statements. After all, the SPAN message is not flight-critical if you don't have it.

Image

src/ubx.cpp
ThomasRigi: header size. The magic 4 was already present in the handling of UBX_MSG_MON_RF, I just copied from there. Ok to create a `static constexpr uint16_t kHeaderSizeMonRFSpan = 4`.

src/ubx.cpp
dakejahl: There are a lot of old modules with old firmware out in the wild, many people never update the firmware. We shouldn't break functionality with these old modules by failing on a NACK for a key that modules firmware doesn't know about.

src/ubx.cpp
dakejahl: automatic_gain_control was recently added for logging purposes. A little bit of duplication with an optional rarely-enabled topic is fine.

src/ubx.cpp
```diff
@@ -1239,6 +1202,23 @@ int GPSDriverUBX::configureDevice(const GPSConfig &config, const int32_t uart2_b
...
+	// UBX_MSG_MON_SPAN
+	cfg_valset_msg_size = initCfgValset();
+	uint8_t value = _spectrum_analyzer ? 1 : 0;
+	cfgValset<uint8_t>(UBX_CFG_KEY_MSGOUT_UBX_MON_SPAN_UART1, value, cfg_valset_msg_size);
+
+	UBX_DEBUG("Configuration spectrum analyzer");
+
+	if (!sendMessage(UBX_MSG_CFG_VALSET, _tx_cfg_valset_buf, cfg_valset_msg_size)) {
+		UBX_WARN("configuring UBX_MSG_MON_SPAN FAILED");
+		return -1;
```
```suggestion
```

src/ubx.cpp
```diff
@@ -1239,6 +1202,23 @@ int GPSDriverUBX::configureDevice(const GPSConfig &config, const int32_t uart2_b
...
+	}
+
+	if (waitForAck(UBX_MSG_CFG_VALSET, UBX_CONFIG_TIMEOUT, true) < 0) {
+		UBX_WARN("configuring UBX_MSG_MON_SPAN FAILED");
+		return -1;
```
```suggestion
```

dakejahl:
needs a rebase

src/ubx.cpp
ThomasRigi: My point was that `configureDevice` is only called when `_proto_ver_27_or_higher`, which won't be the case on any such older non-updated firmware devices.

Anyway, I've reworked it a bit so that we only try to configure when GPS_UBX_SPECTRUM is set: at this point you explicitly ask for it to be set, so I expect it to report a failure. And it doesn't do anything when disabled, so same as now.

src/ubx.h
```diff
@@ -309,7 +311,7 @@
 #define UBX_CFG_KEY_CFG_UART2_STOPBITS          0x20530002
 #define UBX_CFG_KEY_CFG_UART2_DATABITS          0x20530003
 #define UBX_CFG_KEY_CFG_UART2_PARITY            0x20530004
-#define UBX_CFG_KEY_CFG_UART2_ENABLED           0x20530005
+#define UBX_CFG_KEY_CFG_UART2_ENABLED           0x10530005
```
ThomasRigi, line 326: stumbled on this while working with a colleague on a ValGet implementation. I know it's not related to the MR, I can create a dedicated one if prefered ;)

Both https://content.u-blox.com/sites/default/files/documents/u-blox-F9-HPG-1.51_InterfaceDescription_UBXDOC-963802114-13124.pdf?utm_content=UBXDOC-963802114-13124 and https://content.u-blox.com/sites/default/files/documents/u-blox-X20-HPG-2.02_InterfaceDescription_UBXDOC-304424225-19967.pdf have `0x10530005`. 

Image

JonasPerolini: Good catch @ThomasRigi , we can also fix UART1 which has the same issue `UBX_CFG_KEY_CFG_UART1_ENABLED` should be `0x10520006`  (confirmed with the links you've sent above)

dakejahl (approved):
Nice looks good, yeah let's add the fix for UART1_ENABLED too

src/ubx.h
ThomasRigi: fixed. Thanks!

src/ubx.cpp
```diff
@@ -2784,11 +2816,72 @@ GPSDriverUBX::payloadRxDone()
...
 
-		_gps_position->noise_per_ms		= _buf.payload_rx_mon_rf.block[0].noisePerMS;
-		_gps_position->automatic_gain_control	= _buf.payload_rx_mon_rf.block[0].agcCnt;
-		_gps_position->jamming_indicator	= _buf.payload_rx_mon_rf.block[0].jamInd;
-		_gps_position->jamming_state		= _buf.payload_rx_mon_rf.block[0].flags;
+			// Extract all RF metrics
+			for (int i = 0; i < rf_blocks; i++) {
+				sensor_gnss_rf_s gnss_rf{};
+				gnss_rf.timestamp_sample = timestamp_sample;
+				gnss_rf.block_id = _buf.payload_rx_mon_rf.block[i].blockId;;
```
JonasPerolini: double semi colon `;;`

src/ubx.h
```diff
@@ -699,6 +703,9 @@ typedef struct {
 	uint8_t reserved0[56];
 } ubx_payload_rx_mon_hw_deprecated_t;
 
+static constexpr uint8_t kMaxBlocks = 2;  ///< handle up to 2 blocks
```
JonasPerolini: The ZED-X20P has 3 blocks. 

https://content.u-blox.com/sites/default/files/documents/ZED-X20P_IntegrationManual_UBXDOC-963802114-12901.pdf

Image

With the kMaxBlocks = 2, every MON-RF message is dropped so sensor_gps.jamming/noise/AGC fields stay at 0 and no sensor_gnss_rf is published. Previously X20 delivered L1 jamming data via block[0].

I understand the RAM concerns with `kMaxBlocks = 3`, maybe an intermediate solution would be: 

- kMonRfMaxBlocks = 3 → full jamming/noise/AGC coverage on all three X20 bands, no RAM cost.
- kMonSpanMaxBlocks = 2 → save the 272 B; document that on 3-block receivers the L5/E5 spectrum is not published (L1 + L2/E6 only).

src/ubx.h
```diff
@@ -391,6 +393,8 @@
 #define UBX_CFG_KEY_MSGOUT_RTCM_3X_TYPE1230_I2C  0x20910303
 #define UBX_CFG_KEY_MSGOUT_UBX_NAV_TIMEGPS_I2C	 0x20910047
 
+#define UBX_CFG_KEY_MSGOUT_UBX_MON_SPAN_UART1 0x2091038c
```
JonasPerolini, line 411: Should we use `#define UBX_CFG_KEY_MSGOUT_UBX_MON_SPAN_I2C	 0x2091038b` to match the other keys above? So it can go through `cfgValsetPort()`, which derives UART1 (+1), USB (+3) and SPI (+4) for the interface in use

JonasPerolini: We would call: `cfgValsetPort(UBX_CFG_KEY_MSGOUT_UBX_MON_SPAN_I2C, _spectrum_analyzer ? 1 : 0, cfg_valset_msg_size);` instead of `cfgValset<uint8_t>(UBX_CFG_KEY_MSGOUT_UBX_MON_SPAN_UART1, 1, cfg_valset_msg_size);`

JonasPerolini (changes requested):

src/ubx.h
ThomasRigi: Personally I prefer to not activate these big messages on all ports where they're unused, but if there is a usecase where you would want to send it over multiple instances we can change it.

src/ubx.h
dakejahl: My vote is we get all of the data. If RAM becomes an issue we can dynamically allocate only when it's enabled. There are likely other less important things using RAM in the system we can trim.

src/ubx.h
dakejahl: > Personally I prefer to not activate these big messages on all ports where they're unused

I agree

JonasPerolini: Let's keep it as it is, closing thread

src/ubx.h
ThomasRigi: Good catch @JonasPerolini !

I simply increased `kMaxBlocks = 3`. Did I understand it correctly that that's what you meant @dakejahl  ?

Stack size increased from original 2040 to 2750 because the union has become 728 bytes longer (was 92, now 820).

dakejahl: hmm I guess that is a lot, maybe we should allocate these structs only when the feature is enabled

ThomasRigi: Ok. Would we best do that by following the same logic as for satellite_info ? Where the message is split in two parts (essentially the header and the repeated body), with the repeating block ubx_payload_rx_nav_svinfo_part2_t. 

Analogously, having ubx_payload_rx_mon_span_block_t as a separate unit in the buffer that needs to be parsed several times increases the buffer "only" by about 200 bytes (hard to get around that with uint8_t spectrum[256]; that has to live somewhere).

Open for other ideas before I start to implement something.

dakejahl: I'd say it doesn't even need to be split necessarily. Just one allocation that happens only when the feature is enabled.

dakejahl: Maybe @alexcekay has an opinion

alexcekay: For the H7 targets we have quite a lot of RAM, but for the F7 based ones which are still used quite a lot by existing users we already are close to RAM limits. If its a feature that can be turned off via KConfig it is not problematic, as users of older boards just don't get the newest features which for them is an easier to accept deal. What hurts more is if stuff is enabled by default because it pushes older targets over their available RAM.

src/ubx.cpp
```diff
@@ -2848,7 +2848,65 @@ GPSDriverUBX::payloadRxDone()
...
+						gnss_rf.center_frequency = kNominalL1Freq;
+
+					} else if (i == 1) { // L2
+						gnss_rf.center_frequency = kNominalL2Freq;
+
+					} else { // should never be the case
+						gnss_rf.center_frequency = 0;
+					}
+
+				} else if (_board == Board::u_blox9_F9P_L1L5 || _board == Board::u_blox10_L1L5) {
```
ThomasRigi, line 3129: I don't have neither a  `u_blox9_F9P_L1L5` nor a `u_blox10_L1L5` board at hand. From the name I figured it was "safe enough" to include it here. Does anyone have such a unit to explicitly test it ? 


Nor a X20 for that matter, but that's clear from the ublox reference datasheet.

Also, are there any other modules that someone has available to test or that would be obvious enough to fill the data ?

src/ubx.h
ThomasRigi: I have to say that's a bit out of my expertise. I can implement either of the versions, but I don't consider myself the best person to take the decision of which path to choose.

dakejahl: The submodule split makes it a bit annoying to do this properly with KConfig. I'd add a macro that defaults to disable the RF stuff added here. In the PX4 side gps driver we can then override/alias to that macro through KConfig

src/ubx.cpp
dakejahl: I've got an L1/L5 module and an X20 I can test with once this is ready

src/ubx.cpp
```diff
@@ -85,6 +85,10 @@ GPSDriverUBX::GPSDriverUBX(Interface gpsInterface, GPSCallbackPtr callback, void
 	_uart2_baudrate(settings.uart2_baudrate),
 	_ppk_output(settings.ppk_output),
 	_jam_det_sensitivity_hi(settings.jam_det_sensitivity_hi)
+#if defined(CONFIG_GPS_UBX_SPAN)
+,
+	_spectrum_analyzer(settings.spectrum_analyzer)
+#endif
```
ThomasRigi, line 113: this one is a bit ugly with the comma... Probably better to move 
```
#if defined(CONFIG_GPS_UBX_SPAN)
	_spectrum_analyzer(settings.spectrum_analyzer),
#endif
```
above 
`_jam_det_sensitivity_hi(settings.jam_det_sensitivity_hi)`

src/ubx.h
```diff
@@ -115,6 +115,7 @@
 #define UBX_ID_CFG_VALDEL     0x8C
 #define UBX_ID_MON_VER        0x04
 #define UBX_ID_MON_HW         0x09 // deprecated in protocol version >= 27 -> use MON_RF
+#define UBX_ID_MON_SPAN       0x31
```
ThomasRigi, line 121: should I also gate the defines behind CONFIG_GPS_UBX_SPAN ?

ThomasRigi:
I rebased and squashed my commits for a cleaner history. From my tests on an F9P it's all good. @dakejahl I'm happy if you can give it a go on an L1/L5 and X20 receiver. Will report with more testing details in the [main PR](https://github.com/PX4/PX4-Autopilot/pull/28173).

PR #223, opened 2026-08-06
feat(ubx): full UBX_MON_RF support

ThomasRigi:
Log all data from UBX_MON_RF.

Belongs to https://github.com/PX4/PX4-Autopilot/pull/28174

### DIFF
--- a/src/ubx.cpp
+++ b/src/ubx.cpp
@@ -2815,6 +2815,7 @@ GPSDriverUBX::payloadRxDone()
 				gnss_rf.noise_per_ms = _buf.payload_rx_mon_rf.block[i].noisePerMS;
 				gnss_rf.automatic_gain_control = _buf.payload_rx_mon_rf.block[i].agcCnt;
 				gnss_rf.jamming_indicator = _buf.payload_rx_mon_rf.block[i].jamInd;
+				gnss_rf.jamming_state = _buf.payload_rx_mon_rf.block[i].flags;
 				gnss_rf.i_offset = _buf.payload_rx_mon_rf.block[i].ofsI;
 				gnss_rf.i_magnitude = _buf.payload_rx_mon_rf.block[i].magI;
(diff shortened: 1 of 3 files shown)


dakejahl:
If this is part of https://github.com/PX4/PX4-GPSDrivers/pull/222 I'd aggregate it all into one PR

src/ubx.cpp
```diff
@@ -2787,11 +2788,41 @@ GPSDriverUBX::payloadRxDone()
 
 	case UBX_MSG_MON_RF:
 		UBX_TRACE_RXMSG("Rx MON-RF");
+		{
+			// Fill fields for SensorGps
+			// todo: remove them from the message definition, as they're now logged twice ?
+			_gps_position->noise_per_ms		= _buf.payload_rx_mon_rf.block[0].noisePerMS;
+			_gps_position->automatic_gain_control	= _buf.payload_rx_mon_rf.block[0].agcCnt;
+			_gps_position->jamming_indicator	= _buf.payload_rx_mon_rf.block[0].jamInd;
+			_gps_position->jamming_state		= _buf.payload_rx_mon_rf.block[0].flags;
```
ThomasRigi, line 2797: These fields are now logged in a duplicate for ubx receivers.

In our fork we removed them, but that would cause issues in general because we're removing info from drivers . 

- automatic_gain_control is used by no other driver, so we could definitely remove that one
- noise_per_ms is used by femtomes
- jamming_indicator used by multiple
- jamming_state also used by multiple

ThomasRigi:
I'm ok if you want to combine the PRs. I thought it was cleaner to have one per topic. This is for https://github.com/PX4/PX4-Autopilot/pull/28174.

This one is for the MON-RF message, the other for the SPAN message.

dakejahl:
In this case since you're doing both at the same time it will be easier to review if they're combined

ThomasRigi:
Ok, I ported the changes into #222 .

Issue #214, opened 2026-07-10
SBF: heading_accuracy publishes AttCovEuler.Cov_HeadHead (variance, deg²) through a deg→rad conversion — mis-scales EKF2 yaw noise up to ~60×


j-flaherty:
## Describe the bug

`GPSDriverSBF` treats `AttCovEuler.Cov_HeadHead` as an angle:

https://github.com/PX4/PX4-GPSDrivers/blob/52bc6ecef4fb80cc5b7fef12095f65dd3cbe957e/src/sbf.cpp#L709-L711

```cpp
float heading_acc = _buf.payload_att_cov_euler.cov_headhead;
heading_acc *= M_PI_F / 180.0f; // deg to rad, now in range [0, 2pi]
_gps_position->heading_accuracy = heading_acc;
```

Per the mosaic reference guide (AttCovEuler, block 5939), `Cov_HeadHead` is a **variance in degrees²**. `heading_accuracy` is defined as a 1-sigma angle in radians (`SensorGps.msg`). The conversion should be:

```cpp
_gps_position->heading_accuracy = sqrtf(_buf.payload_att_cov_euler.cov_headhead) * M_PI_F / 180.0f;
```

## Observed

On a mosaic-G5 with dual-antenna attitude, logged `heading_accuracy` reaches 6.35 (p95) and 739 (max) during degraded attitude — impossible for an angle in [0, 2pi], consistent with deg² passed through ×π/180. Near Cov = 1 deg² the bug is nearly invisible (1 deg² ≈ 1 deg), which is likely why it hasn't been noticed.

## Impact

EKF2 uses this field as GNSS yaw measurement noise (`R_YAW = sq(fmaxf(yaw_acc, ...))` in `gnss_yaw_control.cpp`). At Cov = 60 deg² the published 1.05 rad vs correct 0.135 rad inflates R_YAW ~60x, de-weighting yaw fusion exactly when heading degrades.

The same code exists in the PX4-Autopilot septentrio module: https://github.com/PX4/PX4-Autopilot/blob/d60fea65d23f87b4f65c111a8e49d9d4fabf94d6/src/drivers/gnss/septentrio/septentrio.cpp#L1405

Happy to open a PR for both.

PR #227, opened 2026-08-18
fix(sbf): convert heading covariance to accuracy

Tartistbz:
## Summary

- Convert SBF `Cov_HeadHead` from a variance in deg^2 to a 1-sigma angle in radians before publishing `heading_accuracy`.
- Use the corrected heading accuracy for the GPS yaw-noise input.

## Validation

- Ubuntu 22.04 / g++ 11.4
- CMake + Ninja build and CTest
- 100% tests passed, 0 tests failed

Addresses part of #214

### DIFF
--- a/src/sbf.cpp
+++ b/src/sbf.cpp
@@ -706,9 +706,10 @@ int GPSDriverSBF::payloadRxDone()
 			int error_aux2 = _buf.payload_att_cov_euler.error_aux2;
 
 			if (error_aux1 == 0 && error_aux2 == 0) {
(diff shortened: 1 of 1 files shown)


dakejahl:
I don't think sbf is even used? Was this tested?

Tartistbz:
You're right. I verified that the repository's current CMake test target does not compile or exercise sbf.cpp, so my build and CTest result did not validate this change. I also could not confirm an active consumer for this driver in the current PX4 codebase. I'll close this PR rather than claim validation I do not have. Thanks for pointing this out.

